PR #22, opened 2019-08-16
Create ro.json

tr1ckz:
Romanian translation


tabarra (changes requested):
what comes before the `||||` is the singular form, and after is the plural one

tr1ckz (commented):
fixed typo

tabarra (commented):
ok

PR #26, opened 2019-08-17
Change bash to sh shell

Epictek:
This allows txAdmin to work on distros without bash, for example alpine.


tabarra:
You are using Docker, right?  
Can we get some more people to report if this will work on their Linux distros before I merge this PR?

Yannick-DG:
Works fine on Debian 8

tabarra:
Can you add a simple bypass via the `--use-bash` parameter?

tabarra (approved):

tabarra:
Will edit locally, will be easier lol

PR #29, opened 2019-08-21
Added multilang support and fixed czech in discord translations

rex2630:
Only css needs to be edited


Yannick-DG:
Wouldn't it be better to have id's and class names in a bit more known names? I see some stuff like usernum and such, guess it's better to keep it just username or hUsername as an identity for host username.

rex2630:
I think it doesnt matter. When u are translating from English strings

Yannick-DG:
http://www.openxrest.com/translatejs/ - Have you tried this plugin?

tabarra:
Decided to close this PR since this would be too hard to maintain.  
Also I don't liked the organization too much. As I said on discord, we would need a way better way to organize the translations into files or namespaces.  
  
In the future, if we were to implement interface translation, it would be better to do it first only on the menu to see how that would work before doing the entire thing.

PR #30, opened 2019-08-24
Added fixed scaleform messages

rex2630:


src/webroutes/fxCommands.js
```diff
@@ -41,7 +41,14 @@ module.exports = async function action(res, req) {
         webUtils.appendLog(req, cmd, context);
         let toResp = await globals.fxRunner.srvCmdBuffer(cmd);
         return sendAlertOutput(res, toResp);
-
+        
+   //==============================================	
+	}else if(action == 'admin_announce'){
+        let cmd = `txaAnnounce "${escape(parameter)}"`;
```
tabarra, line 48: Add permission check:
`if(!ensurePermission('commands.message', res, req)) return false;`

extensions/txAdminClient/resource/sv_main.lua
```diff
@@ -113,6 +113,18 @@ function txaBroadcast(source, args)
     CancelEvent()
 end
 
+-- Announce message to all players
+RegisterCommand("txaAnnounce", function(source, args)
```
tabarra: The `RegisterCommand` should be in line 30-ish, and here just the function.  
Just like the other ones

extensions/txAdminClient/resource/cl_scaleform.lua
```diff
@@ -0,0 +1,89 @@
...
+					AnnouncementScaleform:Dispose()
+					fading = false
+				return 
+			end
+			Wait(0)
+		end
+	end)
+end
+
+RegisterNetEvent("Scaleform:Announce")
```
tabarra, line 31: Not sure about using an event for this instead of a function call or export.  
Check how they work to make sure no one else can trigger this other than txAdmin.  
  
Also, I don't want to use the default event name for scaleform to prevent any conflict issues.

extensions/txAdminClient/resource/cl_scaleform.lua
```diff
@@ -0,0 +1,89 @@
+-- Announce Scaleform to every players
+local announce_title = "Staff announce"  -- Set the title of the announcement here.
+local AnnouncementScaleform
+Scaleform = {}
+local scaleform = {}
+scaleform.__index = scaleform
+
+
+function StartDraw(time)
```
tabarra, line 9: Can I get a good reliable source for these functions?  
I don't want to add to txAdmin a ctrl+c+v of some random code from the forum.  
That's how many fivem scripts get some badly optimized functions.

tabarra (changes requested):

extensions/txAdminClient/resource/cl_scaleform.lua
rex2630: Script was taken from this: https://github.com/IllusiveTea/ScaleformAnnouncememts

extensions/txAdminClient/resource/cl_scaleform.lua
rex2630: As first comment says, i didnt created it all, but it looks ok for me.

tabarra: we can't have it as `Scaleform:Announce`, since the original scaleform code already uses it.

rex2630 (commented):
Its  already done

tabarra:
Not compatible anymore.  
If in the future I decide to add scaleform, I will get back to this PR for reference.

PR #78, opened 2020-03-08
Make it save regardless if the file is too small.

Sampleeeee:
In my server.cfg I removed all the comments that aren't needed. This will let them know that they might of done something wrong if they don't know what they're doing but save anyways


tabarra:
Actually interesting idea.  
Just one problem: People that remove more shit that they are supposed to *probably* aren't reading things anyways,  
  
How about if we do some `cfgData.contains()` to make sure at least some important crap is there like license, hostname and etc then prompt a confirm box listing the missing items?
Just throwing out the idea.

tabarra:
Merged manually.  
I'm always saving an backup file in case of...

PR #83, opened 2020-03-26
Create ar.json - arabic translation

IbrahemAlobeidi:


tabarra:
Thanks for the commit.  
Due to version+file mismatch this will be added manually.

PR #89, opened 2020-03-30
Linux Instruction Clarification

w33ts:
A lot of people have been saying "where is the citizen folder" and such, so I added the full path from the binaries.


Rhydium (approved):
Seems like a good addition to the documentation.

PR #113, opened 2020-04-24
Added Portuguese (Portugal) as an language.

jooaooalmeida:
This was a very minor change as Portuguese (Brazil) and Portuguese (Portugal) are almost the same


jooaooalmeida:
Found some errors, i will fix those lul

jooaooalmeida:
There fixed it lol

tabarra (approved):

PR #127, opened 2020-05-21
Added a basic console history, which can be accessed with the up and down keys

EvanDarwin:
- Added a basic console history, which can be accessed with the up and down keys
- Removed jQuery from the console page, use Vanilla JS instead


w33ts:
😩

LawyerMorty97:
- LilDolla: Does not appear to work.

LawyerMorty97:
JACKBOI: Appears to be working.

PR #134, opened 2020-05-23
Translation from English to Norwegian.

LawyerMorty97:


tabarra:
I changed the singular version of 'minutes' and removed that weird `NB!`.  
Please let me know if I fucked up.

PR #149, opened 2020-06-01
Automatic WL cleanup of inactive players.

SeaquestNet:
Hi!

I added an option to turn on automatic WL cleaning of inactive players (amount of days can be set in settings). We have this functionality on our server in our current whitelisting script but we wanna merge over to txAdmin as soon as possible. 

My current concerns/questions:
- How often to run this, might suffice with once a day?
- Is the "cronjob" section the correct way of running it?
- When removing whitelist from a player, i'm currently revoking the whitelist action. However I am also removing the player object from the database (otherwise the same player would be polled over and over again). Good or bad?

Would you have a look at it and give me some pointers/thoughts?


SeaquestNet:
Added cleanup functionality to checkPlayerJoin function. Left my original code for now but will clean that up once we settle on a final solution.

What do you think?

SeaquestNet:
Did you get a chance to take a look at this?

tabarra:
Unfortunately not yet

SeaquestNet:
Can we solve this by any chance? The server im developing for is looking to merge the whitelist to txAdmin but we can't do this without this feature. 

Thoughts?

locale/fr.json
```diff
@@ -11,6 +11,9 @@
         "schedule_warn": "Le serveur redémarrera dans %{smart_count} minute. Merci de vous déconnecter. |||| Le serveur redémarrera dans %{smart_count} minutes.",
         "schedule_warn_discord": "**%{servername}** redémarrera dans %{smart_count} minute. |||| **%{servername}** redémarrera dans %{smart_count} minutes."
     },
+    "player_actions": {
+        "whitelist_expired": "You have been inactive for too long and your whitelist has expired."
```
rdlh, line 15: `Vous avez été inactif trop longtemps, votre whitelist est arrivée à expiration.`

src/components/playerController/index.js
```diff
@@ -267,6 +277,50 @@ module.exports = class PlayerController {
...
+        }
+
+        try {
+            const revokeDays = this.config.automaticWhitelistCleanupDays ? this.config.automaticWhitelistCleanupDays : 30;
+            const revokeTime = now() - (86400 * +revokeDays);
+            await this.dbo.get('players')
+                .filter(player => player.tsLastConnection <= revokeTime)
+                .value()
+                .forEach(async player => {
+                    let idArray = [];
```
lze3: Can be `const`, you don't reassign its value.
```suggestion
                    const idArray = [];
```

src/components/playerController/index.js
```diff
@@ -408,6 +462,40 @@ module.exports = class PlayerController {
                                     .replace(/<id>/g, `<code>${whitelistID}</code>`);
                     return {allow: false, reason};
                 }
+
+                if(this.config.automaticWhitelistCleanup){
+                    let license = idArray.find((id) => id.substring(0, 8) == "license:").substring(8);
+                    let player = await this.dbo.get('players').find({license: license}).value();
```
lze3: ditto
```suggestion
                    const license = idArray.find((id) => id.substring(0, 8) == "license:").substring(8);
                    const player = await this.dbo.get('players').find({license: license}).value();
```

lze3 (changes requested):

tabarra:
For merging this I was first waiting for the "database cleanup page", and it is [finally done](https://i.imgur.com/bfjjnOP.png).  
But after over a year things changed too much :(  
  
This will still be a thing somewhen.  
Thank you for your PR.

PR #150, opened 2020-06-03
Updated Chinese(Simplified, Traditional) lang

CCasusensa:
chinese simplified and chinese traditional separate


w33ts:
Free Hong Kong.

tabarra (changes requested):
Unfortunately I cannot accept this PR since it will break the config of all existing servers using `zh` as their language.  
Furthermore, I don't feel like this is required, since there are only about 30 chinese servers on the list (compared to about 800 combined for Brazil and Portugal, one that by itself probably shouldn't have been separated into two distinct languages).  
  
Pardon my ignorance, but how close are the two languages?  
If possible, just PR changes to the existing `zh.json` file with the most popular choice (simplified?).

CCasusensa:
sorry, I have merged and change to simplified now

tabarra:
Thank you.

tabarra (approved):

PR #185, opened 2020-08-13
Use systeminformation package to get memory data

### LINKED ISSUE #19: Total memory usage misleading in Linux instances
The current memory usage implementation uses os.freemem() which reports buff/cache as memory used. Avaliable memory should be used instead as it's a lot more useful imo. Maybe even show both as a stacked bar? https://getbootstrap.com/docs/4.3/components/progress/#multiple-bars

Trsak:
This resolves #19 and shows real available memory on Linux.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4958,6 +4958,11 @@
         "has-flag": "^4.0.0"
       }
     },
+    "systeminformation": {
+      "version": "4.26.10",
+      "resolved": "https://registry.npmjs.org/systeminformation/-/systeminformation-4.26.10.tgz",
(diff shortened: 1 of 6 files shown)


tabarra (changes requested):
Module and code looks clean, but this uses WMIC on Windows, so its pretty much unacceptable.   
In my test took full 8 seconds for grabbing this information (remember, `status.js` is called at least once a second).  
  
If you only do this for when `GlobalData.osType === 'linux'`, then I think its fine.  
But still, check the diagnostics page for the cache function, would be good if we could cache this information for at least 30 seconds. FXServer wont make the memory fluctuate that much anyways...

Trsak:
Changed so **systeminformation**  is used only by Linux in bbf7679295df0a5af6a527d0e8dd3bb3551afc1d.
And caching memory data for 30 seconds in e23f58d99e45ba802d0569142683d20f8e0cf193.

Trsak:
Made systeminformation be used only on Linux and added 30 seconds cache for host status data.

tabarra (approved):

PR #186, opened 2020-08-13
Added convar txAdminBindHost for binding specific hostname or IP

### LINKED ISSUE #21: [Discussion] Bind to specific IP
I have looked through everything and I would want to bind this to the specific IP of 127.0.0.1 and then proxy it behind nignx or apache.  Is there a way to do that?

Trsak:
Resolves #21 by adding convar **txAdminBindHost**.  **txAdminBindHost** can be hostname, IPv4 or IPv6 address.

Tested and works.


ablax (approved):
Looks good and needed feature to me

tabarra:
This code is satisfactory, but I chose against this for the reason provided in the comment below:  
https://github.com/tabarra/txAdmin/issues/21#issuecomment-751390616

PR #197, opened 2020-08-28
add zh-CN locale

1107128185:


tabarra:
We decided not to break languages files after our experience with portuguese from Brazil/Portugal.  
If the current translation is not good enough, please edit the current one to try to find some "simplified" phrases that can be understood well by both dialects.

PR #208, opened 2020-09-23
Add txAdmin:shutdown server event.

aultra:
Added a command to trigger a server event just before restart to allow resources a few seconds to save any necessary data.  This command is automatically triggered 5 seconds prior to the server performing the "quit" command.

-- It could be argued that this is redundant, as the fxserver allows for an event just prior to shutdown.  However, I have found in my experience that that event can be too short for some slightly more complicated saves.  Someone manually restarting a server has an opportunity to manually trigger any saves necessary prior to restart.  I have not been able to find any obvious way to detect txAdmin performing a restart (with the exception of the chat message events,) and thought that this might be helpful to coders who want to ensure that the data they need saved, is in fact saved prior to a restart.

### DIFF
--- a/scripts/sv_main.lua
+++ b/scripts/sv_main.lua
@@ -34,6 +34,7 @@ CreateThread(function()
     RegisterCommand("txaBroadcast", txaBroadcast, true)
     RegisterCommand("txaSendDM", txaSendDM, true)
     RegisterCommand("txaReportResources", txaReportResources, true)
+    RegisterCommand("txaShutdown", txaShutdown, true)
     CreateThread(function()
         while true do
             HeartBeat()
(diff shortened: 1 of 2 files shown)


tabarra:
Unfortunately this won't work in general.  
You are hooking it to the server restarter, so it would execute every time, and waiting 5 seconds for any restart is bad.  
  
What we could do is hook it to the schedule restarter where the function would also pass the number of seconds until the restart, so you will start getting events since 30 minutes prior.  
Also this entire part of fxRunner is particularly finicky, so I wouldn't like to throw random 5 seconds delay to it.  
  
**i'm closing this PR, but this issue is still open to discussion, and I'm open to new ideas**

PR #211, opened 2020-10-02
Removed the 16 character limit on CitizenFX id

AleksanderEvensen:
Sorry but this was so irritating while giving permissions to people with long names on the CitizenFX forum


tabarra:
Would take too much time for you to go on fivem and actually see the limit instead of jusr removing this?!
```html
<input name="username" autocomplete="discourse" placeholder="" maxlength="20" id="new-account-username" class="ember-text-field ember-view" type="text">
```  
Please fix it to 20 instead of removing the limit.

PR #221, opened 2020-10-16
Update fxmanifest.lua

aidenzybell:
Update to the latest FXv2 version and add a fancy comment.


tabarra:
not yet :p  
We don't need to raise this requirement.

PR #224, opened 2020-10-25
Added a event for txaLogger to include player name on server side

AleksanderEvensen:
Ehem my friend found out that if you had a CommandHandler on the server-side and triggered the event "txaLogger:CommandExecuted", the player name would not be included in the Server Log. By adding a new event called "txaLogger:CommandExecutedServer" and taking the source as an extra argument this problem was resolved.

Example:
server.lua
`
TriggerEvent('txaLogger:CommandExecutedServer', rawCmd, playerSource)
`

### DIFF
--- a/scripts/sv_logger.js
+++ b/scripts/sv_logger.js
@@ -197,6 +197,14 @@ onNet('txaLogger:CommandExecuted', (data) => {
     }
 });
 
+on('txaLogger:CommandExecutedServer', (data, source) => {
+    try {
+        logger.r(source, 'CommandExecuted', data);
+    } catch (error) {
+        logError(error)
+    }
+});
+
 onNet('txaLogger:DebugMessage', (data) => {
     try {
         logger.r(global.source, 'DebugMessage', data);
--- a/scripts/sv_logger.js
+++ b/scripts/sv_logger.js
@@ -189,17 +189,9 @@ onNet('txaLogger:DeathNotice', (killer, cause) => {
     }
 });
 
-onNet('txaLogger:CommandExecuted', (data) => {
+onNet('txaLogger:CommandExecuted', (data, playerSource) => {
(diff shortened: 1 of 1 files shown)


AleksanderEvensen:
This has been tested

AleksanderEvensen:
Hope you like this change better than the last one, since you didn't like the old name "txaLogger:CommandExecutedServer" i changed the existing one to check if a player Source has been added manually, if not just use the global.source

AleksanderEvensen:
@tabarra have you had some time to look at the code.

tabarra:
Sorry for the delay, due to time constraints all PRs (except dark mode) were delayed to the next update.  
Indeed I had some time to check this, so I'm going to quote here the reply I sent to a friend regarding this:

> The behaviour he is trying to "fix" was very very much not accidental ~~(it's even stated in the extra_logging docs page)~~, txAdmin should only show the command executed from X person if it can verify that was really the source, and on server-side this is not a thing **txAdmin** can do.  
So, for arbitrary messages, we need to create a new logging event where the server can send whatever they want to the log.  
This is not "hard" to do, but was also not a priority.
This also needs a lot of testing lol

So, I haven't decided what to do with this PR yet... not sure if I should merge it manually after some modifications or not.  
Either ways, thank you very much for the PR, and soon we will find out what we are going to do with it.

AleksanderEvensen:
I understand that you want to make sure that the player source is the correct one. Maybe adding a prefix or suffix, explaining that the source might be wrong and this should be taken into consideration when using the custom source argument.

AleksanderEvensen:
Example: 
[10:11:08 PM] %PLAYER_NAME%(not sure) executed: /command Hello World

AleksanderEvensen:
I may re-do this one in the future

Issue #99, opened 2020-04-06
[REQUEST] Option to set restart warning times.


MisterMIntBoy:
Add the ability to change the frequency of notifications for scheduled restarts on Discord. It's annoying to have 30,15,10,5,4,3,2,1 minute reminders. Please allow the ability to change it to what we specify.

tabarra:
What do you want to have, instead of `30,15,10,5,4,3,2,1`?

MisterMIntBoy:
30, 5, and then the actual restart.

Tuwkan618:
maybe 30, 15, 5 and 1

Joelelfstrom:
Since no update has come, I will do it myself. I will do a pull after, should be simple

jooaooalmeida:
> Since no fix has come, I will try fix it myself. I will post here if I manage to fix it, should be simple

How do you fix something that is not a bug????????? It's made that way, it's that way, there is no user friendly way to change it, if u guys are on github I assume u know it has source code and u can do whatever u want if it's inside the license rules... If u want to make it an option u can code it and make a pull request

Also I don't see how its annoying, do you guys keep txadmin logs in #general of your server?

Joelelfstrom:
Was easy to find, but since the files not open for me I can't edit them. Been looking in my phone so far, so will try at home later to do a manual install

https://github.com/tabarra/txAdmin/blob/master/src/components/monitor/index.js

Joelelfstrom:
> > Since no fix has come, I will try fix it myself. I will post here if I manage to fix it, should be simple
> 
> How do you fix something that is not a bug????????? It's made that way, it's that way, there is no user friendly way to change it, if u guys are on github I assume u know it has source code and u can do whatever u want if it's inside the license rules... If u want to make it an option u can code it and make a pull request
> 
> Also I don't see how its annoying, do you guys keep txadmin logs in #general of your server? 

So we have a channel in our discord named Server-Status where it posts restarts. It spams like above-mentioned. 
Since only each client can mute the channel we can't control notifications. So everyone get a notification 10times before restart everytime, YES it's annoying. 

I will do a pull after yes

jooaooalmeida:
> > > Since no fix has come, I will try fix it myself. I will post here if I manage to fix it, should be simple
> > 
[...]

PR #226, opened 2020-10-27
Add Restart warnings to settings and trigger events

### LINKED ISSUE #217: Fire events before restart
Every time, when restart chat warning would show, fire event.

We could then handle this event, show our own message or to other stuff (like save things before restart).

Example:
```
TriggerServerEvent("txAdmin:scheduledRestart", 10)
```

And in our resources:

```
AddEventHandler('txAdmin:scheduledRestart', function(secondsToRestart)
  print("There will be restart in " .. secondsToRestart .. " seconds.")
  
  if secondsToRestart == 10 then
    saveAllUsersToDatabase()
  end
end)
```

Trsak:
Adds option to select how many minutes before restart should be message printed:
Settings

Also triggers event on those times, that can be handled like this in other resources:
```
RegisterNetEvent("txAdmin:events:restart")
AddEventHandler(
    "txAdmin:events:restart",
    function(data)
        if type(source) == "string" or source == nil or source == 0 then
            local remaingTime = tonumber(data.minutesRemaining)
            print("SERVER WILL BE RESTARTED IN " .. remaingTime)
        end
    end
)
```

Resolves #217


Trsak:
Also resolves #99

Sojobo:
Having an event that fires with the restart timer would be greatly appreciated. +1

Trsak:
Any chance this can be looked into?

tabarra:
Will be in this update, don't worry.  
-- I need to finish the server deployer part first.

tabarra:
> Will be in this update, don't worry.
> -- I need to finish the server deployer part first.

Why do I do this to myself? LMAO  
Update must be out in a few hours and I haven't even looked at this....

thephilluk:
Hi,
is this available in the 3.1.0 release or is this still waiting to be published to beta?
thanks,
Philip

PR #237, opened 2020-11-29
Automatically remove /status messages after 30 seconds

Sojobo:
My discord gets spammed by the /status message a lot. Auto cleaning those messages up feels like a good idea to me. Might be cool to have some settings for this but I'll wait to see if this is even wanted before attempting a more involved change.


tabarra:
This PR is full of unrelated changes.  
Hopefully very soon (v3.2?) we will create a persistent status message that will be updated automatically... solving your issue.  
  
ps.: I know that txAdmin is due for a general linting.

PR #238, opened 2020-12-04
Fix build script to work on Unix systems

ekralc:
The build fails on Unix systems as the .cmd file doesn't exist, this solution will work on both Windows and Unix.

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "description": "FiveM FXServer Admin - remotely manage your GTA5 FiveM Server",
   "main": "src/index.js",
   "scripts": {
-    "build": "./node_modules/.bin/webpack.cmd --config webpack.config.js --progress"
+    "build": "webpack --config webpack.config.js --progress"
   },
   "repository": {
     "type": "git",


tabarra:
Hmm, not everybody has `webpack` installed as a global package tho.  
Also, how is the fivem pipeline building it? By running it directly or executing `npm run build`? If they are executing the script, then changing it will likely break the CICD pipeline.

ekralc:
This doesn't require you to have it installed globally as NPM still checks locally https://docs.npmjs.com/cli/v6/configuring-npm/folders#executables - removing the exact path to node_modules is just allowing NPM to do the work, although you could still use `./node_modules/.bin/webpack` and it would work the same.

The FiveM pipeline executes webpack directly without the NPM script so this shouldn't have an effect:

https://github.com/citizenfx/fivem/blob/master/ext/system-resources/build.cmd#L30

https://github.com/citizenfx/fivem/blob/master/ext/system-resources/build.sh#L12

tabarra:
Totally forgot to merge this. thanks @Jamelele!

PR #252, opened 2021-01-03
REDM

tr1ckz:
Added REDM support to fxmanifest and cl_ui.lua

### DIFF
--- a/scripts/cl_ui.lua
+++ b/scripts/cl_ui.lua
@@ -10,9 +10,14 @@ function openWarning(author, reason, tTitle, tWarnedBy, tInstructions)
     Citizen.CreateThread(function()
         local countLimit = 100 --10 seconds
         local count = 0
+        local SpaceKey = 22
+            if GetConvar("gamename", "rdr3") then 
+                SpaceKey = 0xD9D0E1C0
+            end
         while true do
             Citizen.Wait(100)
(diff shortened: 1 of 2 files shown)


scripts/cl_ui.lua
```diff
@@ -10,9 +10,13 @@ function openWarning(author, reason, tTitle, tWarnedBy, tInstructions)
     Citizen.CreateThread(function()
         local countLimit = 100 --10 seconds
         local count = 0
+        local SpaceKey = 22
+        if GetConvar("gamename", "rdr3") then
```
lze3: Perhaps implement something like what the default `chat` resource has for checking if the current environment is RDR2 or not:

https://github.com/citizenfx/cfx-server-data/blob/master/resources/%5Bgameplay%5D/chat/cl_chat.lua#L1

.. also, this condition will always return a truthy value. You probably wanted `GetConvar("gamename", "gta5") == "rdr2"`, but yea, I'd go with `chat`'s way of checking.

lze3 (changes requested):

scripts/cl_ui.lua
tr1ckz: that was good, thanks

tr1ckz:
fixed the redm issue

PR #260, opened 2021-01-11
persian language support

BugCode1:


tabarra:
Partially translated, and looks weird (two alphabets!?).  
Will merge anyways, please be kind and fix it for the next update.

PR #262, opened 2021-01-17
Add github action

Poggicek:
Add automatic building and deploying when you push a commit with a tag "v*". It uses some shitty free github virtual servers so you don't have to pay shit, and it's fast Pog.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,34 @@
+name: Production build
+
+on:
+  push:
+    tags:
+      - v*
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        node-version: [14.x]
+
+    steps:
+      - uses: actions/checkout@v2
+      - name: 'Use Node.js ${{ matrix.node-version }}'
+        uses: actions/setup-node@v1
+        with:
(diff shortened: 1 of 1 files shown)


jooaooalmeida (approved):
Worked during the tests on the fork.

.github/workflows/build.yml
```diff
@@ -0,0 +1,34 @@
...
+  push:
+    tags:
+      - v*
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        node-version: [14.x]
```
tabarra, line 13: needs to be built with node 12, since fxserver uses node 12

.github/workflows/build.yml
```diff
@@ -0,0 +1,34 @@
...
+
+      - run: npm i
+      - run: npm run build
+      - uses: papeloto/action-zip@v1
+        with:
+          files: dist/
+          recursive: true
+          dest: monitor.zip
+
+      - uses: "marvinpinto/action-automatic-releases@latest"
```
tabarra, line 30: this repo seems kinda small for something that is super common... perhaps there is a reason for that?  
May be a good idea to try to find another one.

Poggicek:
going to experiment with different release script

PR #263, opened 2021-01-18
Add automatic production build

Poggicek:
Automatic build & deploy. Using and [create-release](https://github.com/actions/create-release) and [upload-release-asset](https://github.com/actions/upload-release-asset). This utilizes github actions virtual machine so you don't have to self host a worker. The average build & deploy time from push is around 30 seconds.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,49 @@
+name: Production build
+
+on:
+  push:
+    tags:
+      - v*
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        node-version: [12.x]
+
+    steps:
+      - uses: actions/checkout@v2
+      - name: 'Use Node.js ${{ matrix.node-version }}'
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -0,0 +1,49 @@
...
+
+    steps:
+      - uses: actions/checkout@v2
+      - name: 'Use Node.js ${{ matrix.node-version }}'
+        uses: actions/setup-node@v1
+        with:
+          node-version: '${{ matrix.node-version }}'
+
+      - name: Download all modules
+        run: npm i
```
TasoOneAsia: Run `ci` instead of `i` to ensure clean dependencies
```suggestion
        run: npm ci
```

TasoOneAsia (commented):
otherwise lgtm

Poggicek:
npm i would also do the job just fine as the vm automatically deletes all the files including node_modules everytime it downloads the repo

tabarra:
merged already

PR #271, opened 2021-01-26
Fix: Killer player is incorrect

rdlh:
Hi there 👋 !

We're used to have incorrect killers in our logs and managed to fix it by using the `GetPlayerServerId` native function.
This piece of code can (maybe) be replaced with `GetPlayerServerId(killerPed)`, but we didn't tested it yet.

Some explanations can be found [here](https://forum.cfx.re/t/fivem-core-why-the-players-id-is-different-on-the-client-and-on-the-server/12128/4).

Cheers and keep the good work !


AvarianKnight:
Wouldn't this also need `GetPlayerFromIndex` removed on the server side code as it expects a index, not a source.
https://github.com/tabarra/txAdmin/blob/115c37420d4f82e39de3c290c4361bcec34efe46/scripts/sv_logger.js#L178

This would also break if the ped wasn't a player?
https://github.com/tabarra/txAdmin/blob/115c37420d4f82e39de3c290c4361bcec34efe46/scripts/cl_logger.lua#L51

scripts/cl_logger.lua
```diff
@@ -39,11 +39,11 @@ local function processDeath(ped)
         end
     else
         if IsEntityAPed(killerPed) and IsPedAPlayer(killerPed) then
-            killer = NetworkGetPlayerIndexFromPed(killerPed)
+            killer = GetPlayerServerId(NetworkGetPlayerIndexFromPed(killerPed))
```
AvarianKnight: would probably be cleaner to add onto the if statement on line 57 instead of adding it to each `NetworkGetPlayerIndexFromPed`
```lua
if killer == nil then
	killer = false
elseif killer then
	-- we have their player, we need their server id
	killer = GetPlayerServerId(killer)
end
```

scripts/cl_logger.lua
```diff
@@ -39,11 +39,11 @@ local function processDeath(ped)
...
         elseif IsEntityAVehicle(killerPed) then
             local drivingPed = GetPedInVehicleSeat(killerPed, -1)
             if IsEntityAPed(drivingPed) == 1 and IsPedAPlayer(drivingPed) then
-                killer = NetworkGetPlayerIndexFromPed(drivingPed)
+                killer = GetPlayerServerId(NetworkGetPlayerIndexFromPed(drivingPed))
             else
                 killer = false
```
AvarianKnight, line 175: line 51 needs to be changed to `killer = false` instead of `killer = killerPed`

AvarianKnight (changes requested):

AvarianKnight:
This would need to not use `GetPlayerFromIndex`, as we're already getting the source + the fact that it expects a integer (usually 0-GetPlayers()), not sure how your code even worked with this set how it is.
https://github.com/tabarra/txAdmin/blob/115c37420d4f82e39de3c290c4361bcec34efe46/scripts/sv_logger.js#L178

rdlh:
Hi @AvarianKnight !

Sorry, I forget to paste this part of our ugly monkey patch. This should be ok now.

AvarianKnight (approved):
This should solve all of the issues, tested with 150 players and it seems to be logging correctly.

image

rdlh:
@AvarianKnight Confirmed, tested on a 320 / 330 players RP server 👌.

rdlh:
Will work on some irrelevant `unknown` causes later this week. 

Possibly caused by missing hash in cl_logger.lua line 4.

Will open another PR.

AvarianKnight:
A lot of them seem to just be missing hashes.

rdlh:
Hi @AvarianKnight, added missing hashed and improved readability in my 613dc2d commit.

tabarra (commented):
Shouldn't we use the backtick here?
Also, if we actually do have the reason, probably doesn't make sense to group it... what's the best way to translate the hashvalue of `WEAPON_PISTOL_MK2` to `PISTOL MK2`? Maybe just remove `WEAPON_` then replace `_` with spaces.

AvarianKnight:
> Shouldn't we use the backtick here?
> Also, if we actually do have the reason, probably doesn't make sense to group it... what's the best way to translate the hashvalue of `WEAPON_PISTOL_MK2` to `PISTOL MK2`? Maybe just remove `WEAPON_` then replace `_` with spaces.

Main issue is that backticks break intellisense for a few extensions (like https://marketplace.visualstudio.com/items?itemName=sumneko.lua) plus its only used once to get the hash so it shouldn't effect performance.

rdlh:
@tabarra Hi mate, I think the best way to do this is just to create a hash with weapon name as the key, and the translated name as the value.
This will also allow people to translate it if needed.

tabarra (changes requested):
Nice job guys.  
But at this point, since we already have all the hash keys, what if we put them all initially in an array, and at start time a loop would populate a permutation table containing the hash as the key, and the weapon tag as value with the `WEAPON_` prefix removed.  
Need to release an update _now_, otherwise I would do it myself.

tabarra:
Hey guys, any hope of implementing what I asked above?  
In short, wince we now have the hash names, we could give the ACTUAL cause of death instead of the generic categories I used previously.

PR #274, opened 2021-01-28
feat(sv_logs): convert the logger to lua and use fd3

AvarianKnight:


PLOKMJNB (approved):

scripts/sv_logs.lua
```diff
@@ -0,0 +1,111 @@
+if GetConvar('txAdminServerMode', 'false') == 'true' then
```
tabarra: Hmm, use this instead:
```lua
if GetConvar('txAdminServerMode', 'false') ~= 'true' then
    return
end
```

scripts/sv_logs.lua
```diff
@@ -0,0 +1,111 @@
...
+	local function getPlayerData(src)
+		if not src then return false end
+		if src == -1 then return {name = 'console', identifiers = {}} end
+		return {
+			name = GetPlayerName(src),
+			identifiers = GetPlayerIdentifiers(src)
+		}
+	end
+
+	local apiPort = GetConvar("txAdmin-apiPort", "invalid")
```
tabarra: throw api port and token part to the top of the file, like the sv_main.lua

src/components/fxRunner/outputHandler.js
```diff
@@ -98,10 +98,12 @@ module.exports = class OutputHandler {
...
         //Handle script traces
         if(channel == 'citizen-server-impl' && data.type == 'script_structured_trace'){
-            // dir(data.payload)
+			// dir(data.payload)
             if(data.payload.type === 'txAdminHeartBeat'){
                 globals.monitor.handleHeartBeat('fd3');
-            }
+            } else if (data.payload.type === 'txAdminLogData') {
+				globals.databus.serverLog = globals.databus.serverLog.concat(data.payload.logs)
+			}
```
tabarra: fix mixed whitespaces

tabarra: Also in general, make sure you are only using 4 spaces instead of tabs.  
So I guess also fix the other files.

scripts/sv_logs.lua
```diff
@@ -0,0 +1,111 @@
...
+	local apiToken = GetConvar("txAdmin-apiToken", "invalid")
+	if apiPort == "invalid" or apiToken == "invalid" then
+		logError('API Port and Token ConVars not found. Do not start this resource if not using txAdmin.')
```
tabarra: I don't think logError was declared, does it carry from the other file? 🤔

scripts/sv_logs.lua
```diff
@@ -0,0 +1,111 @@
...
+		return
+	end
+
+	-- back ticks break my intellisense
+	local PRINT_STRUCTURED_TRACE = GetHashKey('PRINT_STRUCTURED_TRACE')
+	local function log(src, action, data)
```
tabarra: `log` and `logError` words should be used for terminal logging, like in `sv_main.lua`.
Probably better to make those print functions a separate file and include here.  
  
So, for the actual fd3 log function, use another name.  
Also to prevent spamming and broken pipes, add all to a buffer and flush it every second or so. Maybe even 2~3.

tabarra (changes requested):

scripts/sv_logs.lua
AvarianKnight: Yes, its something that carrys over from sv_main, didn't realize that log does too 😨

scripts/sv_logs.lua
AvarianKnight: > So, for the actual fd3 log function, use another name.
> Also to prevent spamming and broken pipes, add all to a buffer and flush it every second or so. Maybe even 2~3.

Would this really be necessary? It was something that was in the initial concept when it was initially using HTTP instead of fd3, but it complicated the code a lot and I didn't see it as something worth having, though this could be due to my lack of knowledge of how txAdmin and fd3 in general works.

AvarianKnight:
If it all looks good to you I'll add the explosions later https://github.com/citizenfx/fivem-docs/pull/238

PR #275, opened 2021-01-29
Add a broadcast event with the ability to turn off legacy announcement prints

nnsdev:
I added an event for admin broadcasts under `txAdmin:events:broadcast` for custom styling abilities.

I went by most of how the original PR regarding server restarts has been created, but I am uncertain about the location of the toggle as well as pulling in the config in `fxserver/commands.js`. Listening to feedback regarding that, willing to move it.

### DIFF
--- a/docs/events.md
+++ b/docs/events.md
@@ -25,3 +25,16 @@ AddEventHandler('txAdmin:events:scheduledRestart', function(eventData)
     end
 end)
 ```
+
+## txAdmin:events:broadcast
+Called for whenever an announcement is triggered from the admin panel.  
+Arguments:
+- `author`: Username of the person triggering the announcement.  
+- `announcement`: Announcement text
+
+Example usage:
+```lua
+AddEventHandler('txAdmin:events:broadcast', function(eventData)
+    print(("announcement %s: %s"):format(eventData.author, eventData.announcement))
+end)
+```
--- a/src/webroutes/fxserver/commands.js
+++ b/src/webroutes/fxserver/commands.js
@@ -68,10 +68,16 @@ module.exports = async function FXServerCommands(ctx) {
     //==============================================
     }else if(action == 'admin_broadcast'){
         if(!ensurePermission(ctx, 'players.message')) return false;
-        let cmd = formatCommand('txaBroadcast', ctx.session.auth.username, parameter);
+        const config = globals.fxRunner.config
+        const cmd = formatCommand('txaBroadcast', ctx.session.auth.username, parameter);
+        const eventCmd = formatCommand('txaEvent', 'broadcast', JSON.stringify({author: ctx.session.auth.username, announcement: parameter}));
+        globals.fxRunner.srvCmdBuffer(eventCmd);
         ctx.utils.logCommand(cmd);
-        let toResp = await globals.fxRunner.srvCmdBuffer(cmd);
-        return sendAlertOutput(ctx, toResp);
+        if (config.disableAnnouncementPrint !== 'true' && config.disableAnnouncementPrint !== true) {
+            let toResp = await globals.fxRunner.srvCmdBuffer(cmd);
+            return sendAlertOutput(ctx, toResp)
+        }
+        return sendAlertOutput(ctx, cmd)
 
     //==============================================
     }else if(action == 'kick_all'){
--- a/web/settings.html
+++ b/web/settings.html
@@ -194,6 +194,23 @@
                             </div>
                         </div>
 
+                        <div class="form-group row">
+                            <label class="col-sm-3 col-form-label">
+                                Disable Announcement Warnings
+                            </label>
+                            <div class="col-sm-9">
+                                <label class="c-switch c-switch-label c-switch-pill c-switch-success fix-pill-form">
+                                    <input class="c-switch-input" type="checkbox" id="frmFXServer-announcementPrint"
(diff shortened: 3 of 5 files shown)


docs/events.md
```diff
@@ -25,3 +25,16 @@ AddEventHandler('txAdmin:events:scheduledRestart', function(eventData)
     end
 end)
 ```
+
+## txAdmin:events:broadcast
+Called for whenever an announcement is triggered from the admin panel.  
+Arguments:
+- `user`: Username of the person triggering the announcement.
```
tabarra: use "author" instead

docs/events.md
```diff
@@ -25,3 +25,16 @@ AddEventHandler('txAdmin:events:scheduledRestart', function(eventData)
...
 end)
 ```
+
+## txAdmin:events:broadcast
+Called for whenever an announcement is triggered from the admin panel.  
+Arguments:
+- `user`: Username of the person triggering the announcement.  
+- `announcement`: Announcement text
+
+Example usage on ESX v1.2:
```
tabarra: this code doesn't depend on esx at all

src/webroutes/fxserver/commands.js
```diff
@@ -68,10 +68,16 @@ module.exports = async function FXServerCommands(ctx) {
...
-        ctx.utils.logCommand(cmd);
-        let toResp = await globals.fxRunner.srvCmdBuffer(cmd);
-        return sendAlertOutput(ctx, toResp);
+        const config = globals.configVault.getScoped('monitor')
+        const cmd = formatCommand('txaBroadcast', ctx.session.auth.username, parameter);
+        const eventCmd = formatCommand('txaEvent', 'broadcast', JSON.stringify({user: ctx.session.auth.username, announcement: parameter}));
+        globals.fxRunner.srvCmdBuffer(eventCmd);
+        if (config.disableAnnouncementPrint !== 'true' && config.disableAnnouncementPrint !== true) {
+            let toResp = await globals.fxRunner.srvCmdBuffer(cmd);
+            ctx.utils.logCommand(cmd);
```
tabarra: The execution should be logged even if the chat message is disabled

src/webroutes/fxserver/commands.js
```diff
@@ -68,10 +68,16 @@ module.exports = async function FXServerCommands(ctx) {
     //==============================================
     }else if(action == 'admin_broadcast'){
         if(!ensurePermission(ctx, 'players.message')) return false;
-        let cmd = formatCommand('txaBroadcast', ctx.session.auth.username, parameter);
-        ctx.utils.logCommand(cmd);
-        let toResp = await globals.fxRunner.srvCmdBuffer(cmd);
-        return sendAlertOutput(ctx, toResp);
+        const config = globals.configVault.getScoped('monitor')
```
tabarra: In txAdmin we would usually do `globals.fxRunner.config.xxxxxx` instead of calling the config vault.
i know this is not ideal, but that's how our scoped settings work.

web/settings.html
```diff
@@ -246,6 +246,25 @@ <h2>Settings</h2>
                             </div>
                         </div>
 
+
```
tabarra: No double line break, follow the pattern from the other switches.

tabarra (changes requested):
Thanks for the commit.
Please review the pending changes I set, and move this setting to the FXServer Settings tab instead of Monitor (same goes for the scope of the config file).

docs/events.md
nnsdev: Whoops, copy pasted this from the other event documentation, forgot to remove it lol.

nnsdev:
👍 currently busy with work, I'll have the PR updated by wednesday or thursday

nnsdev:
Requested changes have been resolved

tabarra:
So, with the anniversary update incoming, we need to solve this PR.  
The `txaBroadcast` command will be removed, as well as all chat messages (maybe will just let DMs), and txAdmin will have an NUI to display announcements, which will use the `txaEvent` itself.  
  
Now what we need to find out is how we gonna signal txAdmin's NUI to not act on the event.  
Should we wait for a `CancelEvent()` or add a `disableNUI: true` to the event, so  the NUI knows it should ignore it?  
@TasoOneAsia any input?  
  
BTW, we are now using [globals.fxRunner.sendEvent()](https://github.com/tabarra/txAdmin/blob/feat/menu/src/components/fxRunner/index.js#L370) (in the `feat/menu` branch).

nnsdev:
Do you want me to rebase to `feat/menu`?

tabarra:
I think its probably a good idea, and point the PR to the `feat/menu` branch.

TasoOneAsia:
Maybe a CancelEvent() could do the trick but we could also get context with the web panel setting setting a convar. Then we can update the ServerCtx and cancel the event that way.

tabarra (changes requested):
I talked to Taso, and the best thing would be:
1. Use [globals.fxRunner.sendEvent()](https://github.com/tabarra/txAdmin/blob/feat/menu/src/components/fxRunner/index.js#L370) for the events;
2. Move this settings to the global scope. Asking for it to move to fxserver was a mistake. My bad;
3. Use the setting to set a `showNUI` parameter, which will inform txAdmin NUI if it should display the announcement or not.
  
I already pointed the PR to `feat/menu`.
Thanks for all the changes you did to this PR =]

TasoOneAsia:
Stale PR, can reopen if updated

PR #277, opened 2021-01-30
Bind to localhost

ablax:
Add option to bind to localhost to be able to create nginx cached reverse proxy in front of it.


tabarra:
Hey thanks for the PR, but this is an old issue that we already have a verdict on.  
If you don't want to expose txadmin to the internet, better to just not open the firewall port for it.  
References:
#186 and https://github.com/tabarra/txAdmin/issues/21#issuecomment-751390616

PR #289, opened 2021-02-21
Update de.json

NoobKeksTV:
replace a word  to make it more "normalized"  / make it more understandable

Made ban messages similar to each other


tabarra:
Kick reasons cannot contain line break.... sadly!

PR #291, opened 2021-02-23
Chat message changed for Admin DM and broadcast

GamenHaaDee:


tabarra:
Interesting, when was this template thing added?  
txAdmin must be compatible with older chat resources.

GamenHaaDee:
If im right its added like a year or 2 ago

tabarra:
Can you find me the commit? Also a screenshot of what it looks like the template?

GamenHaaDee:
yeah if u have a second

GamenHaaDee:
image

image

GamenHaaDee:
https://github.com/citizenfx/cfx-server-data/pull/1
it goes back to 2017

tabarra:
I just tested and it looks okay, but I see how people would strongly dislike it.  
Currently we already have a lot of people asking for a way to customize the messages/announcements to be more close to their server's style, and adding something strongly "opinionated" like that would just cause unnecessary conflict.  
  
I'm closing this because I believe the best course of action here would be to adapt the warn page to also do announcements (WAY SMALLER, way less intrusive, likely blue). Perhaps that's something you would like to try?! 😉

PR #293, opened 2021-02-25
added server restart & stop events to be fired to fivem side

NoobKeksTV:


Arm-0001:
Doesn't work: 
hamy-is-Knowing

NoobKeksTV:
fixxed & tested now. working for me

tabarra:
I didn't add these events on purpose.  
They are called 100ms before server restarts, which give the server virtually no time to do anything with those events.  
  
People will try to plug their `esx.save()` and them complain it does not work.

NoobKeksTV:
normally, esx does auto saves on disconnecting the players. so it should be a problem normally??

but tstill server should be able to do it in time even with 200 players (at least mine does lmao) 

you could put a bigger wait before tho :)

PR #311, opened 2021-03-19
Update fx_version

LiamDormon:
Updated FX version from 'bodacious' to 'cerulean'. Bodacious is deprecated and cerulean is the latest


tabarra:
Although we don't *need* this, and `bodacious` not actually being "deprecated", I see no reason why not update since txAdmin can only run in versions that support `cerulean`.

PR #312, opened 2021-03-20
Tweak(css) Changed dark mode to a more dimmed "chilled out" version

Arm-0001:
dark mode looks so sexy now https://beta.iodine.gg/8c71e.png https://beta.iodine.gg/i44ib.png https://beta.iodine.gg/0jdje.png


tabarra:
1. I don't like it
2. Nobody is asking for it
3. Changing just two colors isn't enough, there are a bunch of hue changes required

PR #335, opened 2021-05-02
updated source on getPlayerData

synterrr:
source -1 doesn't exists, i tried using a command chat on the terminal "say something" and wasn't working, now it works👍

### DIFF
--- a/scripts/sv_logger.lua
+++ b/scripts/sv_logger.lua
@@ -14,7 +14,7 @@ local function getPlayerData(src)
     if not src then
         return false
     end
-    if src == -1 then return {name = 'console', identifiers = {}} end
+    if src == 0 then return {name = 'console', identifiers = {}} end
 
     return {
         name = GetPlayerName(src),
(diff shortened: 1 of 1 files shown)


PLOKMJNB (approved):

Nimplex (approved):

scripts/sv_logger.lua
```diff
@@ -14,7 +14,7 @@ local function getPlayerData(src)
     if not src then
         return false
     end
-    if src == 0 then return {name = 'console', identifiers = {}} end
+    if src <= 0 then return {name = 'console', identifiers = {}} end
```
TasoOneAsia, line 17: This less than is not necessary. Should be fine as ==

TasoOneAsia (changes requested):

scripts/sv_logger.lua
synterrr: ok

BombayV (approved):

PR #336, opened 2021-05-03
Updated polish translation

Nimplex:


tabarra (changes requested):
Thanks for the PR.  
Do you mind updating the `restarter` part? We changed the texts and keys.

tabarra:
Thanks! Will merge it soon!

PR #339, opened 2021-05-08
Update tr.json

abdulkadiraktas:


tabarra (changes requested):
A "hang" or "freeze" is not the same thing as a "crash", that's why I changed the english message.  
And "hang" is not the same thing as a "partial".

abdulkadiraktas:
I see. I fixed.

locale/tr.json
```diff
@@ -6,9 +6,9 @@
     "restarter": {
         "start_timeout": "zamanında başlatılamadı",
         "crash_detected": "Çökme tespit edildi",
-        "hang_detected": "hang detected",
-        "partial_hang_warn": "Due to a partial hang, this server will restart in 1 minute. Please disconnect now.",
-        "partial_hang_warn_discord": "Due to a partial hang, **%{servername}** will restart in 1 minute.",
+        "hang_detected": "Kısmi bir takılma algılandı",
```
tabarra: Does "Kısmi" mean Partial? If that's the case then its better to remove the work, not to confuse with the partial hang warnings.

abdulkadiraktas: "hang" has a lot of means from Turkish. In This case "Takılma" is good. I deleted "Kısmi" side.
http://prntscr.com/12q28zd

tabarra:
Thanks! Will merge it soon!

PR #359, opened 2021-05-30
DO NOT MERGE: Pull request to hold feat/menu gameplay impact review

blattersturm:
As @tabarra requested review of @GoatG33k's gameplay logic, here's a draft PR so I can use GitHub's review UI to apply inline comments.

Note that my review here primarily applies to gameplay code, i.e. the scripts/menu/ directory.


scripts/menu/client/cl_menu.lua
```diff
@@ -0,0 +1,554 @@
...
+  isMenuDebug = (GetConvar('TXADMIN_MENU_DEBUG', 'false') == 'true')
+end)
+
+-- Since the menu yields/receives keyboard
+-- focus we need to store that the menu is already visible
+local isMenuVisible
+-- Last location stored in a vec3
+local lastTp
+
+RegisterKeyMapping('txadmin', 'Open the txAdmin Menu', 'keyboard', '')
```
blattersturm: Uncertain: does passing `''` as preferred key correctly register the binding in the pause menu binding options? If not, that's a GAMETODO.

scripts/menu/client/cl_menu.lua
```diff
@@ -0,0 +1,554 @@
...
+function sendMenuMessage(action, data)
+  SendNUIMessage({
+    action = action,
+    data = data
+  })
+end
+
+--- Will update ServerCtx based on GlobalState and will send it to NUI
+local function updateServerCtx()
+  local ServerCtx = GlobalState.txAdminServerCtx
```
blattersturm, line 80: State will likely not exist when state awareness is disabled. Since this mode is _extremely_ deprecated for GTA5, that might be a good thing.

scripts/menu/client/cl_menu.lua
```diff
@@ -0,0 +1,554 @@
+-- Variable that determines whether a player can even access the menu
```
blattersturm, line 2: Add a return here so this code does not run on RDR, perhaps.

scripts/menu/client/cl_menu.lua
```diff
@@ -0,0 +1,554 @@
...
+  if menuIsAccessible then
+    toggleMenuVisibility()
+  else
+    sendSnackbarMessage('error', 'nui_menu.misc.menu_not_allowed', true)
+  end
+end)
+
+-- alias
+RegisterCommand('tx', function()
+  ExecuteCommand('txadmin')
```
blattersturm: Does this work? (I'd suspect so, due to the default RegisterCommand permissions allowing `builting.everyone` and there being no mandatory access control)

scripts/menu/client/cl_menu.lua
```diff
@@ -0,0 +1,554 @@
...
+end)
+
+CreateThread(function()
+  TriggerEvent('chat:addSuggestion', '/txadmin', 'Open the txAdmin menu', {})
```
blattersturm: These suggestions are useless to players who don't have access. I'd rather opt to _remove_ the auto-added suggestion from chat's resource start handler, and add them once it is known if the player has access. to the menu.

scripts/menu/client/cl_menu.lua
```diff
@@ -0,0 +1,554 @@
...
+  TriggerEvent('chat:addSuggestion', '/tx', 'Open the txAdmin menu', {})
+  SetNuiFocus(false, false)
```
blattersturm: Doesn't make much sense, resource stop already should clear focus and mouse focus.

scripts/menu/client/cl_menu.lua
```diff
@@ -0,0 +1,554 @@
...
+  elseif mode == 'none' then
+    toggleGodMode(false)
+    toggleFreecam(false)
+  end
+end)
+
+RegisterNUICallback('spawnWeapon', function(weapon, cb)
+  debugPrint("Spawning weapon: " .. weapon)
+  local playerPed = PlayerPedId()
+  local weaponHash = GetHashKey(weapon)
```
blattersturm: I *believe* GetHashKey for these (and following uses) is inferred if the native has `Hash` type and you pass a `string` value.

scripts/menu/client/cl_menu.lua
```diff
@@ -0,0 +1,554 @@
...
+  GiveWeaponToPed(playerPed, weaponHash, 500, false, true)
+  cb({})
+end)
+
+-- CB From Menu
+RegisterNUICallback('spawnVehicle', function(data, cb)
+  local model = data.model
+  if not IsModelValid(model) then
```
blattersturm: I'd append to this a check for `IsModelAVehicle` in case the input here is directly user-provided. Also, if possible, use `IsThisModelACar` to specify whether the server can use the `CREATE_AUTOMOBILE` fast-path or has to defer to RPC-based `CREATE_VEHICLE` + awaiting creation.

scripts/menu/client/cl_menu.lua
```diff
@@ -0,0 +1,554 @@
...
+  if (veh == 0) then
+    return cb({ e = true })
+  end
+
+  TriggerServerEvent('txAdmin:menu:fixVehicle')
+  cb({})
+end)
+
+-- Used to trigger the help alert
+AddEventHandler('playerSpawned', function()
```
blattersturm, line 372: This event will trigger _every_ spawn, which may lead to a fun overlap of a lot of 'helpful message' timers and a bit of spam if the user spawns a lot.

In addition to that, consider storing a client resource-keyed KVP value with when you've last shown this 'helpful message' to a given player to reduce annoyance.

scripts/menu/client/cl_menu.lua
```diff
@@ -0,0 +1,554 @@
...
+  end)
+end
+
+--[[ Player list sync ]]
+RegisterNetEvent('txAdmin:menu:setPlayerState', function(data)
+  -- process data to add distance, remove pos
+  for i in ipairs(data) do
+    local row = data[i]
+    local targetVec = vec3(row.pos.x, row.pos.y, row.pos.z)
+    local dist = #(GetEntityCoords(PlayerPedId()) - targetVec)
```
blattersturm: `GetEntityCoords(PlayerPedId())` is constant for the entire duration of this loop.

scripts/menu/client/cl_menu.lua
```diff
@@ -0,0 +1,554 @@
...
+--[[ Teleport the player to the coordinates ]]
+---@param x number
+---@param y number
+---@param z number
+RegisterNetEvent('txAdmin:menu:tpToCoords', function(x, y, z)
+  local ped = PlayerPedId()
+  lastTp = GetEntityCoords(ped)
+  
+  DoScreenFadeOut(200)
+  while not IsScreenFadedOut() do Wait(1) end
```
blattersturm: `Wait(1)` is meaningless and generally equal to `Wait(0)`.

scripts/menu/client/cl_menu.lua
```diff
@@ -0,0 +1,554 @@
...
+    data = data
+  })
+end)
+
+--[[ Teleport the player to the coordinates ]]
+---@param x number
+---@param y number
+---@param z number
+RegisterNetEvent('txAdmin:menu:tpToCoords', function(x, y, z)
+  local ped = PlayerPedId()
```
blattersturm, line 484: Hypothetically, while waiting, the ped model may change invalidating this handle.

scripts/menu/client/cl_menu.lua
```diff
@@ -0,0 +1,554 @@
...
+  lastTp = GetEntityCoords(ped)
+  
+  DoScreenFadeOut(200)
+  while not IsScreenFadedOut() do Wait(1) end
+  
+  debugPrint('Teleporting to coords')
+  if z == 0 then
+    for i = 0, 1000, 10 do
```
blattersturm: Painful loop that may lead to a 3-4-second delay at worst. I'd rather repeatedly process a vertical line downwards as with the following pseudocode:

```
z := 1500
found := true

repeat
    found, z = getGroundZ x y (z - 1)
    wait 0
until not found or z <= 0
```

This way you'll still get the lowest ground coordinate available, while avoiding a 100-frame worst-case wait if moving around a high place.

scripts/menu/client/cl_menu.lua
```diff
@@ -0,0 +1,554 @@
...
+      local zFound, _z = GetGroundZFor_3dCoord(x + 0.0, y + 0.0, i + 0.0)
+      if zFound then
+        debugPrint("3D ground found: " .. json.encode({ zFound, _z }))
+        z = _z
+        break
+      end
+      Wait(0)
+    end
+  end
+  RequestCollisionAtCoord(x, y, z)
```
blattersturm: Useless in itself as next box streamer tick will mark as unneeded, and this is asynchronous. Since you're also moving the ped (and have moved it a few ticks ago, perhaps), the box streamer will in fact _already_ request it.

scripts/menu/client/cl_menu.lua
```diff
@@ -0,0 +1,554 @@
...
+  RequestAdditionalCollisionAtCoord(x, y, z)
```
blattersturm: And this is outright harmful! Unless you (or the game?) clear it shortly after this will retain these coords as an additional position for the box streamer to keep loaded for all eternity.

scripts/menu/client/cl_menu.lua
```diff
@@ -0,0 +1,554 @@
...
+    
+    DoScreenFadeOut(200)
+    while not IsScreenFadedOut() do Wait(1) end
+    
+    local blipCoords = GetBlipInfoIdCoord(waypoint)
+    debugPrint("waypoint blip: " .. json.encode(blipCoords))
+    local x = blipCoords[1]
+    local y = blipCoords[2]
+    local z = blipCoords[3]
+    for i = 0, 1000, 10 do
```
blattersturm: Code duplication, and the comments above still apply.

scripts/menu/client/cl_menu.lua
```diff
@@ -0,0 +1,554 @@
...
+
+-- [[ Teleport to the current waypoint ]]
+RegisterNetEvent('txAdmin:menu:tpToWaypoint', function()
+  local waypoint = GetFirstBlipInfoId(GetWaypointBlipEnumId())
+  if waypoint and waypoint > 0 then
+    local ped = PlayerPedId()
+    lastTp = GetEntityCoords(ped)
+    
+    DoScreenFadeOut(200)
+    while not IsScreenFadedOut() do Wait(1) end
```
blattersturm: Another `Wait(1)` antipattern.

scripts/menu/client/cl_menu.lua
```diff
@@ -0,0 +1,554 @@
...
+  end
+end)
+
+--[[ Heal all players ]]
+RegisterNetEvent('txAdmin:menu:healed', function()
+  debugPrint('Received heal event, healing to full')
+  local ped = PlayerPedId()
+  local pos = GetEntityCoords(ped)
+  if IsEntityDead(ped) then
+    ResurrectPed(ped)
```
blattersturm: Is this valid for a network player? The use of `SET_ENTITY_COORDS` below and the fact you're dealing with the local ped makes it more likely that you should use the resurrect-local-network-player function.

scripts/menu/client/cl_menu.lua
```diff
@@ -0,0 +1,554 @@
...
+    SetVehicleEngineOn(veh, true, false)
+    SetVehicleDirtLevel(veh, 0.0)
+    SetVehicleOnGroundProperly(veh)
+  end
+end)
+
+--[[ Spawn vehicles, with support for entity lockdown ]]
+RegisterNetEvent('txAdmin:menu:spawnVehicle', function(netID)
+  debugPrint(json.encode({ netID = netID }))
+  SetNetworkIdExistsOnAllMachines(netID, true)
```
blattersturm: If the network ID doesn't exist yet, this is nonsense.

In fact, this call and the below call are somewhat nonsense already, even in non-state-awareness mode!

scripts/menu/client/cl_menu.lua
```diff
@@ -0,0 +1,554 @@
...
+  
+  -- get current veh and speed
+  local ped = PlayerPedId()
+  local oldVeh = GetVehiclePedIsIn(ped, false)
+  local oldVel = DoesEntityExist(oldVeh) and GetEntitySpeed(oldVeh) or 0.0
+  
+  -- only delete if the new vehicle is found
+  if oldVeh and IsPedInVehicle(ped, oldVeh, true) then
+    debugPrint("Deleting existing vehicle (" .. oldVeh .. ")")
+    SetEntityAsMissionEntity(oldVeh, true, true)
```
blattersturm: Not needed in Cfx since we don't do script ownership checks on deletion calls. Also I believe this won't even work if the entity is held by another script.

scripts/menu/client/cl_menu.lua
```diff
@@ -0,0 +1,554 @@
...
+    DeleteVehicle(oldVeh)
+  end
+  
+  local tries = 0
+  while NetworkGetEntityFromNetworkId(netID) == 0 do
```
blattersturm: This'll log a warning if it doesn't exist - see NetworkDoesEntityExistWithNetworkId instead.

scripts/menu/client/cl_menu.lua
```diff
@@ -0,0 +1,554 @@
...
+    end
+    debugPrint("Waiting for vehicle networking...")
+    tries = tries + 1
+    Wait(50)
+  end
+  
+  local veh = NetworkGetEntityFromNetworkId(netID)
+  debugPrint("Found networked vehicle " .. netID .. " (entity " .. veh .. ")")
+  SetVehicleHasBeenOwnedByPlayer(veh, true)
+  TaskWarpPedIntoVehicle(ped, veh, -1)
```
blattersturm: SetPedIntoVehicle... which also exists as a server RPC native?

scripts/menu/client/cl_menu.lua
```diff
@@ -0,0 +1,554 @@
...
+  SetEntityHeading(veh, GetEntityHeading(ped))
```
blattersturm: Both server-side vehicle spawn commands actually take an initial heading now, I believe. However, given deferred creation, sure, correction _may_ make sense.

scripts/menu/server/sv_menu.lua
```diff
@@ -0,0 +1,405 @@
...
+    headers['X-TxAdmin-Identifiers'] = table.concat(GetPlayerIdentifiers(s), ', ')
+  else
+    headers['X-TxAdmin-Token'] = 'not_required' -- so it's easy to detect webpipes
+  end
+  
+  local url = "http://" .. apiHost .. path:gsub("//", "/")
+  debugPrint(string.format("^3WebPipe[^5%d^0:^1%d^3]^0 ^4>>^0 ^6%s^0", s, callbackId, url))
+  debugPrint(string.format("^3WebPipe[^5%d^0:^1%d^3]^0 ^4>>^0 ^6Headers: %s^0", s, callbackId, json.encode(headers)))
+
+  PerformHttpRequest(url, function(httpCode, data, resultHeaders)
```
blattersturm, line 150: Potential for some nasty abuse from event handler submission here. I'd stick to only allowing `/auth/nui` for players that aren't in the `adminPermissions` table, and caching a failure or attempt to call it way too often, and performing a no-op otherwise.

scripts/menu/server/sv_menu.lua
```diff
@@ -0,0 +1,405 @@
...
+CreateThread(function()
+  -- If we don't wait for a tick there is some race condition that
+  -- sometimes prevents debugPrint lmao
+  Wait(0)
+  syncServerCtx()
+end)
+
+RegisterServerEvent('txAdmin:menu:getServerCtx', function()
+  local src = source
+  TriggerClientEvent('txAdmin:menu:sendServerCtx', src, serverCtxObj)
```
blattersturm: Is this still used? Seems like a great way to get oneself spammed with events to reject and clog up the server send list otherwise.

scripts/menu/server/sv_menu.lua
```diff
@@ -0,0 +1,405 @@
...
+  local allow = PlayerHasTxPermission(src, 'players.heal')
+  TriggerEvent("txaLogger:menuEvent", src, "healSelf", allow)
+  if allow then
+    TriggerClientEvent('txAdmin:menu:healed', src)
+  end
+end)
+
+RegisterServerEvent('txAdmin:menu:healPlayer', function(id)
+  local src = source
+  local parseId = type(id) == 'string' and tonumber(id) or id
```
blattersturm: If type is not string or number, you're in for a bad time if your admin is in a bad mood.

scripts/menu/server/sv_menu.lua
```diff
@@ -0,0 +1,405 @@
...
+  if not parseId then
+    return
+  end
+  local allow = PlayerHasTxPermission(src, 'players.heal')
+  local playerName = GetPlayerName(id) or "unknown"
```
blattersturm: s/your admin/anyone/ as now you're passing this unchecked value to a native no matter if the user has access lol

scripts/menu/server/sv_menu.lua
```diff
@@ -0,0 +1,405 @@
...
+RegisterServerEvent('txAdmin:menu:tpToPlayer', function(id)
+  local src = source
+  local parseId = type(id) == 'string' and tonumber(id) or id
+  if not parseId then
+    return
+  end
+
+  local allow = PlayerHasTxPermission(src, 'players.teleport')
+  local data = { x = nil, y = nil, z = nil, playerName = nil }
+  data.playerName = GetPlayerName(parseId)
```
blattersturm: Same checking concern here, what if the user sends a table? lol

scripts/menu/server/sv_menu.lua
```diff
@@ -0,0 +1,405 @@
...
+      data.z = coords[3]
+      TriggerClientEvent('txAdmin:menu:tpToCoords', src, data.x, data.y, data.z)
+    end
+  end
+  TriggerEvent('txaLogger:menuEvent', src, 'teleportPlayer', allow, data)
+end)
+
+RegisterServerEvent('txAdmin:menu:summonPlayer', function(id)
+  local src = source
+  local parseId = type(id) == 'string' and tonumber(id) or id
```
blattersturm: Parsing logic duplication yet again and also unconditional GetPlayerName.

scripts/menu/server/sv_menu.lua
```diff
@@ -0,0 +1,405 @@
...
+  local playerName = GetPlayerName(parseId)
+  TriggerEvent('txaLogger:menuEvent', src, 'summonPlayer', allow, playerName)
+end)
+
+RegisterServerEvent('txAdmin:menu:tpToWaypoint', function()
+  local src = source
+  local allow = PlayerHasTxPermission(src, 'players.teleport')
+  if allow then
+    TriggerClientEvent('txAdmin:menu:tpToWaypoint', src)
+    Wait(250)
```
blattersturm, line 387: A bit optimistic, innit?

scripts/menu/server/sv_menu.lua
```diff
@@ -0,0 +1,405 @@
...
+      { x = coords[1], y = coords[2], z = coords[3] })
+  else
+    TriggerEvent("txaLogger:menuEvent", src, "teleportWaypoint", false)
+  end
+end)
+
+RegisterServerEvent('txAdmin:menu:sendAnnouncement', function(message)
+  local src = source
+  local allow = PlayerHasTxPermission(src, 'players.message')
+  TriggerEvent("txaLogger:menuEvent", src, "announcement", allow, message)
```
blattersturm, line 400: now your logger may get some huge table, hf

scripts/menu/server/sv_menu.lua
```diff
@@ -0,0 +1,405 @@
...
+RegisterServerEvent('txAdmin:menu:fixVehicle', function()
+  local src = source
+  local allow = PlayerHasTxPermission(src, 'menu.vehicle')
+  TriggerEvent("txaLogger:menuEvent", src, "vehicleRepair", allow)
+  if allow then
+    TriggerClientEvent('txAdmin:menu:fixVehicle', src)
+  end
+end)
+
+local CREATE_AUTOMOBILE = GetHashKey("CREATE_AUTOMOBILE")
```
blattersturm: `.

Also, see client comment: you should not use CREATE_AUTOMOBILE on an arbitrary model index as it may be another vehicle type leading to an inconsistent net clone.

Note, yet again, that CREATE_VEHICLE is a bit deferred and won't get a network ID until a client created it (i.e. DOES_ENTITY_EXIST loop on server).

scripts/menu/server/sv_menu.lua
```diff
@@ -0,0 +1,405 @@
...
+RegisterServerEvent('txAdmin:menu:spawnVehicle', function(model)
+  local src = source
+  if type(model) ~= 'string' then
+    error()
```
blattersturm: how to get a console to spam lua errors in 2 easy steps:

1. spam this with a non-string
2. there is no step 2

scripts/menu/server/sv_menu.lua
```diff
@@ -0,0 +1,405 @@
...
+      }
+      
+      -- Lets wait a tick so we don't have hitch issues
+      Wait(0)
+    end
+  
+    -- get the list of all players to send to
+    debugPrint("^4Sending ^3" .. #found .. "^4 users details to ^3" .. #adminPermissions .. "^4 admins^0")
+    for id, _ in pairs(adminPermissions) do
+      TriggerClientEvent('txAdmin:menu:setPlayerState', id, found)
```
blattersturm: That's a _lot_ of data for a reliable event, extremely optimistic towards admins having good network connectivity that won't drop, say, 700 players x ~80 bytes per player every 5 seconds without requiring resends.

Latent events would help a bit, along with a lowered rate and _maybe_ eventually keeping some sort of delta (not resending names, e.g.) as events do tend to be _somewhat_ ordered and mainly eventually-consistent.

scripts/menu/server/sv_menu.lua
```diff
@@ -0,0 +1,405 @@
...
+      local veh = GetVehiclePedIsIn(ped)
+      if veh and veh > 0 then
+        veh = NetworkGetNetworkIdFromEntity(veh)
+      else
+        veh = nil
+      end
+  
+      local health = math.ceil(((GetEntityHealth(ped) - 100) / 100) * 100)
+      found[#found + 1] = {
+        id = serverID,
```
blattersturm: Keys are still serialized as key, which is a _slight_ waste of bandwidth too here.

scripts/sv_main.lua
```diff
@@ -14,7 +14,7 @@ end
 local apiHost = GetConvar("txAdmin-apiHost", "invalid")
 local apiToken = GetConvar("txAdmin-apiToken", "invalid")
 local txAdminClientVersion = GetResourceMetadata(GetCurrentResourceName(), 'version')
-if GetConvar('txAdminServerMode', 'false') ~= 'true' then
+if GetConvar('txAdmin-serverMode', 'false') ~= 'true' then
```
blattersturm: This being renamed seems bad, game code checks for this in at least one place to enable certain Node.js behavior fixing issues with JS resources on the server abusing stdout/stdin/stderr.

scripts/menu/client/cl_menu.lua
GoatG33k: It does, but I was being lazy here. I will make them share a handler function.

scripts/menu/client/cl_menu.lua
tabarra: Regarding the annoyance, we are using CEF's localStorage and showing the message only once a day.

tabarra:
(Note: this PR was automatically marked as "merged" by GitHub)

PR #369, opened 2021-06-10
chore(locale): Add Menu Locale

aidenzybell:


TasoOneAsia:
Locales aren’t final yet which is why we haven’t added them in yet

PR #371, opened 2021-06-12
fix: only register key mappings on user auth

GoatG33k:


tabarra (approved):

scripts/menu/client/cl_menu.lua
```diff
@@ -69,7 +69,14 @@ local isMenuVisible
 -- Last location stored in a vec3
 local lastTp
 
-RegisterKeyMapping('txadmin', 'Open the txAdmin Menu', 'keyboard', '')
+function registerTxKeybinds()
+  -- Only register keybinds for authed users
+  if menuIsAccessible then
+    RegisterKeyMapping('txadmin', 'Open the txAdmin Menu', 'keyboard', '')
+    RegisterKeyMapping('txAdmin:menu:endSpectate', 'Exit spectate mode', 'keyboard', 'BACK')
```
blattersturm, line 76: It seems like the registration at the top is still there even with this patch. 😛

GoatG33k: Looks like I forgot to commit that file, fixed with [commit `cf0b3e`](https://github.com/tabarra/txAdmin/commit/cf0b2ee8f049b30299a2991a4ffbac358ab2e00e#diff-4f0ffc4b98c51454b6e949bc1661c385a2499aadb525d9da07010870b14e72c9L12) by tabarra

PR #372, opened 2021-06-15
Added min & max char count to confirm pass

craxydev:


tabarra:
Thanks for the PR!  
This is the confirmation password, and in this case is better if we don't restrict the max input.  
Let's say the dude copies from a password manager a 26 characters password, then he will end up inserting two 24 character passwords that will NOT be the one he thinks he is inserting.  
  
I'm removing the maxlength for the change password modal.

craxydev:
The issue you described is percisely why i made the PR, because i copied a 40 char password from my password manager which got shortened on one input but not the other

tabarra:
Yes, but we have to choose from one of the below:
- Shorten both and the password will be accepted but the admin won't know what the pw is;
- Shorten only one, and then complain that it doesn't match.  

Maybe the best option would be to accept everything and then in the JS message we complain about the size?  
That could be a better option.

PR #384, opened 2021-06-29
fix(menu): freeze player after teleport until the collision has loaded

iQuadCore:
This prevents falling underground if player has slow HDD / low-end PC

### DIFF
--- a/scripts/menu/client/cl_main_page.lua
+++ b/scripts/menu/client/cl_main_page.lua
@@ -172,6 +172,11 @@ RegisterNetEvent('txAdmin:menu:tpToCoords', function(x, y, z)
         if _z ~= nil then z = _z end
     end
     SetPedCoordsKeepVehicle(ped, x, y, z)
+    if not HasCollisionLoadedAroundEntity(ped) then
+        FreezeEntityPosition(ped, true)
+        while not HasCollisionLoadedAroundEntity(ped) do Wait(50) end
+        FreezeEntityPosition(ped, false)
+    end
     DoScreenFadeIn(500)
     SetGameplayCamRelativeHeading(0)
 end)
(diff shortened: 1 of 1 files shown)


tabarra:
Thanks for the PR!  
For me it didn't work, apparently `FindZForCoords()` is always returning 0.

iQuadCore:
@tabarra `FindZForCoords()` returning 0 could be another issue, suspect it happens when teleporting to waypoint. 
The way I tested this is I entered a server with my main PC and with a low-end laptop (~20 FPS in FiveM on minimal settings), and summoned laptop player to main PC player via txAdmin menu, players were 9km away from each other.
Laptop player always fell underground without this changes, and waiting for collisions to load fixed the issue 100%.

tabarra:
Hmm... I'm testing right now and apparently for the tpToWaypoint `HasCollisionLoadedAroundEntity()` returns 1 even when it didn't load, so I'm really at a "wtf" stage.   
Can you try to solve this issue? What worked for me was:
- teleporting the player to the `x, z, 0` coords;
- waiting 500;
- getting the new Z;
- teleporting player again, but with the new Z;

iQuadCore:
Teleporting to z = 0 doesn't work for me in case waypoint is set to high places (e.g. mount Chiliad).

I've also tried teleporting to z = 1000, then decreasing z by 50 and teleporting again, until correct z coords are returned by `FindZForCoords()` function and player can be teleported to precise terrain height. This did work better for high and low places, still not in 100% cases however. If no better solution is found, I can commit this one. _Guess we would have to TP player to z = 1000 and equip him with parachute, when z coord wasn't found after all this manipulations 🙄._

tabarra:
For your code, shouldn't you first check if the player is in the veh, and then freeze the veh instead of the ped?  
Apparently `FreezeEntityPosition()` doesn't work if you freeze the pe and he's in a vehicle.  
(also one of the identation is with tabs instead of spaces)

TasoOneAsia:
LGTM logic wise and worked well in testing, merged into development branch

PR #416, opened 2021-07-30
feat: Add 'Delete Vehicle' option to admin menu

LiamDormon:
This turns the vehicle spawner into a multiple choice giving the option to delete as well as spawn.

### DIFF
--- a/locale/en.json
+++ b/locale/en.json
@@ -95,7 +95,8 @@
                 "dialog_desc": "Spawn a weapon using its model name."
             },
             "car_options": {
-                "list_primary": "Vehicle Options"
+                "list_primary": "Vehicle Options",
+                "onesync_error": "This option requires that OneSync be on for server entity handling"
             },
             "spawn_veh": {
                 "list_primary": "Spawn Vehicle",
@@ -104,14 +105,12 @@
                 "dialog_desc": "Enter in the model name of the vehicle you want to spawn",
                 "dialog_success": "Vehicle spawned!",
                 "dialog_error": "The vehicle model name '%{modelName}' does not exist!",
-                "dialog_info": "Trying to spawn %{modelName}",
-                "onesync_error": "This option requires that OneSync be on for server side spawning"
+                "dialog_info": "Trying to spawn %{modelName}"
             },
             "delete_veh": {
                 "list_secondary": "Delete your vehicle",
                 "dialog_success": "Vehicle Deleted!",
-                "dialog_error": "You are not currently in a vehicle!",
-                "onesync_error": "This option requires that OneSync be on for server entity handling"
+                "dialog_error": "You are not currently in a vehicle!"
             },
             "heal_all": {
                 "list_primary": "Heal All Players",
--- a/locale/en.json
+++ b/locale/en.json
@@ -94,6 +94,9 @@
                 "dialog_title": "Spawn Weapon",
                 "dialog_desc": "Spawn a weapon using its model name."
             },
+            "car_options": {
+                "list_primary": "Vehicle Options"
+            },
             "spawn_veh": {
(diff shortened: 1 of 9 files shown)


locale/en.json
```diff
@@ -104,6 +107,11 @@
                 "dialog_info": "Trying to spawn %{modelName}",
                 "onesync_error": "This option requires that OneSync be on for server side spawning"
             },
+            "delete_veh": {
+                "list_secondary": "Delete your vehicle",
+                "dialog_success": "Vehicle Deleted!",
+                "dialog_error": "You are not currently in a vehicle!"
```
Poggicek, line 113: ~~"You are currently not in a vehicle!" sounds better. I might be wrong, correct me if so. I'm not a native english speaker~~ Nvm. I've said to myself too many times and now it loses it's meaning to me. Looking at stats, "not currently" is used way more

menu/src/state/vehiclemode.state.ts
```diff
@@ -0,0 +1,13 @@
...
+  SPAWN = "spawn",
+  DELETE = "delete",
+}
+
+const vehicleMode = atom({
+  key: "vehicleModeState",
+  default: VehicleMode.SPAWN,
+});
+
+export const useVehicleMode = () => useRecoilState(vehicleMode);
```
TasoOneAsia, line 13: We don't really need this to be a remembered state probably. Fine to just be reset on mount

menu/src/components/MainPage/MainPageList.tsx
```diff
@@ -225,6 +227,23 @@ export const MainPageList: React.FC = () => {
...
+          }
+        );
+      }
+
+      enqueueSnackbar(t("nui_menu.page_main.delete_veh.dialog_success"), {
+        variant: "info",
+      });
+    });
+  }
+
```
TasoOneAsia: We need to have a oneSync ctx check here as we are using server side game state awareness

TasoOneAsia (changes requested):

menu/src/state/vehiclemode.state.ts
LiamDormon: This is how other multi-choice options are mananged, healmode, teleportmode and playermode so I kept that the same for this.

TasoOneAsia: Yea we don't need that for those either, I implemented it with player mode as that is a persistent mode for players, the other ones are somewhat unnecessary.

menu/src/components/MainPage/MainPageList.tsx
LiamDormon: Pushed a commit for this

menu/src/components/MainPage/MainPageList.tsx
```diff
@@ -228,6 +228,11 @@ export const MainPageList: React.FC = () => {
   };
 
   const handleDeleteVehicle = () => {
+    // If onesync is disabled, show an error due to server side entity handling
+    if (!serverCtx.oneSync.status) {
+      return enqueueSnackbar(t('nui_menu.page_main.delete_veh.onesync_error'), { variant: 'error' })
```
TasoOneAsia: Lets just have a combined locale key for both this and spawn vehicle as they are practically the same.

TasoOneAsia (changes requested):

menu/src/components/MainPage/MainPageList.tsx
LiamDormon: Pushed a commit combining them into one locale.

TasoOneAsia:
Rebase against `feat/menu` and squash commits and this should be good, pending @tabarra review as well.

TasoOneAsia (approved):

PR #437, opened 2021-09-06
feat: Added a Convar check to disable 'txadmin:selfDataReport'

NenoxAG:
I've added a Convar check to disable the ``txadmin:selfDataReport`` event being sent from the client to the server.
This is so that you can simply disable that event using ``setr disableSelfDataReport true`` if it bothers you.

### DIFF
--- a/scripts/cl_main.lua
+++ b/scripts/cl_main.lua
@@ -39,25 +39,29 @@ end)
 
 
 -- Self data reporting thread
-CreateThread(function()
-    local ceil = math.ceil
-    
-    while true do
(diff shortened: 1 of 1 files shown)


tabarra:
This entire thing will actually be removed.  
I'll not close the PR for now but i don't think i'll merge it.

tabarra:
Thanks for the PR @NenoxAG!
Since this code was entirely disabled (code commented-out), I'll be closing this PR.  
In some future update we will add a better alternative to that event.

PR #439, opened 2021-09-09
Migrate Menu to Material UI v5

TasoOneAsia:
As Material UI RC v5 has recently been released, we should begin to migrate towards it. There is one dependency that does not yet support Material UI v5 officially (notistack), so this PR may be updated to use that when available.

Aside from that dependency, this migration should go off without a hitch.


TasoOneAsia:
Material UI v5 has been officially released and is ready for production. This pull request has been updated from RC deps to the production deps.

From my testing, this migration hasn't resulted in any unintentional styling or functionality changes and *should* be ready for merge.

@tabarra Am I good to merge into develop?

tabarra:
retracted: notistack fixed :)

PR #440, opened 2021-09-09
feat(web): show server profile in document title

### LINKED ISSUE #405: Suggestion: Put Server Profile Name in the tab label
Pretty much what the title says. I dev for a community with multiple servers and I have all of the different txAdmins open at the same time. It would be nice to see the Profile name in the tab label so I know what server I'm looking at at a quick glance instead of this
image

TasoOneAsia:
May be wise to combine server-name and profile-name in the future for consistency sake & UX. 

**Preview**

img

Closes #405

### DIFF
--- a/src/webroutes/status.js
+++ b/src/webroutes/status.js
@@ -124,7 +124,7 @@ function prepareMetaData() {
     return {
         favicon,
         title: (globals.monitor.currentStatus == 'ONLINE')
-            ? `(${globals.playerController.activePlayers.length}) txAdmin`
-            : 'txAdmin',
+            ? `(${globals.playerController.activePlayers.length}) ${globals.info.serverProfile} | txAdmin`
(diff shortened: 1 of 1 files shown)


tabarra:
I've been thinking and this design is probably bad.  
The status endpoint should return something that is platform independent.  
So for instance it would return the profile name and status, and the presentation layer (web js) would translate that into title and `favicon_xxx`.

TasoOneAsia:
How do you see the favicon being used to convey the specific profile of an instance? It most likely can't be based on color as color is already being used to convey the FXServer status. Do you see it being set by the user for each profile?

tabarra:
Ohh not what I meant lol.  
I was trying to say that the backend should just say if it's online or not, and leave for the frontend to decide what to do with that info... particularly the favicon.

tabarra:
Closed by 3f024df0ba03db55d70e4fed5b930d8118f37d41

PR #455, opened 2021-09-30
Spectating with bucket

Dev-Shaz:


scripts/menu/server/sv_player_modal.lua
```diff
@@ -68,9 +68,25 @@ RegisterNetEvent('txAdmin:menu:spectatePlayer', function(id)
     if not target then
       return
     end
+    local tgtBucket = GetPlayerRoutingBucket(id)
+    local srcBucket = GetPlayerRoutingBucket(src)
+    if tgtBucket ~= srcBucket then 
+      ADMIN_DATA[tostring(src)].bucket = srcBucket
```
TasoOneAsia, line 74: I feel like this isn't the best place to store the data regarding bucket as we haven't traditionally used this map for non-admin related data but meh.

An alternative is we can store the original bucket on the calling client.

scripts/menu/server/sv_player_modal.lua
```diff
@@ -68,9 +68,25 @@ RegisterNetEvent('txAdmin:menu:spectatePlayer', function(id)
...
 
     local tgtCoords = GetEntityCoords(target)
     TriggerClientEvent('txAdmin:menu:specPlayerResp', src, id, tgtCoords)
   end
   TriggerEvent('txaLogger:menuEvent', src, 'spectatePlayer', allow, id)
+end)
+
+
+
+RegisterNetEvent('txAdmin:menu:endSpectate', function()
```
TasoOneAsia, line 86: This net event can be used to spam errors or change an admins bucket without any verification. Please run verification checks on the calling source using the HasPermission function and `players.spectate`

TasoOneAsia (changes requested):

scripts/menu/server/sv_player_modal.lua
Dev-Shaz: Will do my bad

Dev-Shaz: Done

TasoOneAsia (approved):

PR #457, opened 2021-10-02
Add: /addwl <mention>

Dev-Shaz:


tabarra:
Thanks for the PR, but I don't think this one we can merge.  
The players db identifier saving should not work like that. What if an license joins but the discord ID is different?  
There are so many thing to consider...  
  
I postponed those features until the database is saving the entire history of player identifiers, and for me to do that I will also need to redo quite a few UI stuff and web routes.

Dev-Shaz:
no problem

Issue #463, opened 2021-10-10
[BUG] Freeze command does not work for players in vehicles


tabarra:
**Expected behavior**
Should stop the car/vehicle, and on unfreeze, should unfreeze vehicle/player.  
The issue is that a player can exit the vehicle while frozen, so we would need to freeze the player, and freeze te vehicle, but then if he leaves and we unfreeze him, we would also need to unfreeze the vehicle but he's not inside of it anymore... so probably need to keep watching with a thread to make sure we are only freezing a vehicle if he is the driver... idk.

TasoOneAsia:
Fixed in #466

PR #464, opened 2021-10-10
[BUG] Fixed Player Freeze in Vehicle

tim-it-ger:
- added disable 'f' if a player is in a vehicle. to prevent the player to leave the vehicle.

fix for issue: #463

### DIFF
--- a/scripts/menu/client/cl_freeze.lua
+++ b/scripts/menu/client/cl_freeze.lua
@@ -27,6 +27,9 @@ end)
 
 RegisterNetEvent('txAdmin:menu:freezePlayer', function(isFrozen)
   debugPrint('Frozen: ' .. tostring(isFrozen))
+	if IsPedInVehicle(PlayerPedId()) then
+     DisableControlAction(0, 23, isFrozen)
+  end
   FreezeEntityPosition(PlayerPedId(), isFrozen)
   sendFreezeAlert(isFrozen)
 end)


ghost:
You can bypass this by setting another key as leave vehicle or use an controller

scripts/menu/client/cl_freeze.lua
```diff
@@ -27,6 +27,9 @@ end)
 
 RegisterNetEvent('txAdmin:menu:freezePlayer', function(isFrozen)
   debugPrint('Frozen: ' .. tostring(isFrozen))
+	if IsPedInVehicle(PlayerPedId()) then
+     DisableControlAction(0, 23, isFrozen)
```
LiamDormon, line 31: This native would need to be ran every frame.

tim-it-ger: if the player use a Mod Menu yes.
but 1 execution is enough

NoobKeksTV: Instead of blocking the key, just disable movement for Ped & car hes in, if hes the driver. 

This way he cant move at all. He can exit, yes. but cant move.

NoobKeksTV (changes requested):
instead of permanently disablying control, just FreezeEntity of PEd & car if hes the driver.  this way he cant move at all. he can exit but not move.

TasoOneAsia:
Thanks for the PR but the logic present in #466 makes more sense for this use case.

PR #466, opened 2021-10-10
fix(client/freeze): exit vehicle

Jaccosf:
- In my opinion, the best way to do this is let the player leave the vehicle (the vehicle will also stop, if the driver is being freezed) and freeze it after.

### DIFF
--- a/scripts/menu/client/cl_freeze.lua
+++ b/scripts/menu/client/cl_freeze.lua
@@ -29,7 +29,7 @@ RegisterNetEvent('txAdmin:menu:freezePlayer', function(isFrozen)
   debugPrint('Frozen: ' .. tostring(isFrozen))
   local playerPed = PlayerPedId()
   if IsPedInAnyVehicle(playerPed) then
-    TaskLeaveAnyVehicle(playerPed)
+    TaskLeaveAnyVehicle(playerPed, 0, 16)
   end
   FreezeEntityPosition(playerPed, isFrozen)
   sendFreezeAlert(isFrozen)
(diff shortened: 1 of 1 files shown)


Jaccosf:
This will be a fix for: #463

NoobKeksTV:
yeah, this should work. 
#464 is not the right way, it would have to be run in a thread to be called every frame but this makes the ped leave any vehicle hes in so we can freeze him properly. 
Also we could just freeze the Player and the car both but as mentioned by tab in #463 that will require a check for the car when player gets unfreezed.

This gets approval by NoobKeks Pro dev 👍

Jaccosf:
Haha thanks! 😃

scripts/menu/client/cl_freeze.lua
```diff
@@ -27,6 +27,10 @@ end)
 
 RegisterNetEvent('txAdmin:menu:freezePlayer', function(isFrozen)
   debugPrint('Frozen: ' .. tostring(isFrozen))
-  FreezeEntityPosition(PlayerPedId(), isFrozen)
+  local playerPed = PlayerPedId()
+  if IsPedInAnyVehicle(playerPed) then
+    TaskLeaveAnyVehicle(playerPed)
```
TasoOneAsia: Can we pass a flag of `16` to the optional flags please. This flag specifically specifies teleport outside of the vehicle immediately.

[Ref](https://docs.fivem.net/natives/?_0xD3DBCE61A490BE02)

TasoOneAsia (changes requested):

scripts/menu/client/cl_freeze.lua
Jaccosf: Yeah sure

TasoOneAsia:
I think the overall logic is this implementation is sound, I'll slate it for testing tomorrow, thanks for the patience.

PR #468, opened 2021-10-11
Fix Routing Problem when Spectating

NoobKeksTV:
now, you WILL be set to the target bucket no matter what & it will be saved as your old one.

Also cause it will always be set before spectating, theres no need  to reset my saved old bucket back to 0, cause it always get set to the one, i was in.

### DIFF
--- a/scripts/menu/server/sv_player_modal.lua
+++ b/scripts/menu/server/sv_player_modal.lua
@@ -8,6 +8,7 @@ end
 --  related to actions defined within Menu's
 --  "Player Modal"
 -- =============================================
+local oldBucket = 0
 
 RegisterNetEvent('txAdmin:menu:tpToPlayer', function(id)
   local src = source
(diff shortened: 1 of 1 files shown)


TasoOneAsia:
Next time you file a random PR, please state what problem this exactly fixes. I assume this PR is supposed to tackle an issue where the default routing bucket isn't 0?

NoobKeksTV:
tab asked me to do this. i though he knew whtas up.

This pr fixxes the issue that your "old" bucket wont be set to the one ur in, if ur in the same bucket as the player ur spectating . 
BUT You will always get set back to your "old" bucket no matter what. Due to the bucket not being set before, you will always be send to bucket 0, which in some cases is not the right one, you were in.  

for example, my default bucket is bucket 129 to kinda protect from modders & co.  by making them not see anyone untill fully registered. 

My commit fixxes this so your "old" bucket will always be set to the one you were in so you can be returned correctly.

TasoOneAsia:
On my todo list for refactoring, tomorrow.

PR #482, opened 2021-11-02
fix(playerlist): health percentages work as intended

carlos55ml:
this pr is wrong. ignore it.


TasoOneAsia:
At author's request

PR #483, opened 2021-11-02
fix(playerlist): health percentages now work as intended

carlos55ml:
using GetPlayerMaxHealth() seems to fix 38% problem in playerlist... consistently. Also parenthesis have been added.


tabarra:
Note after testing:  
The health shows correctly according to the UI, but when we heal, the health goes to 133% when the old code used to show 100%.  
  
This needs to be investigated a bit more.

PR #485, opened 2021-11-05
Update README.md - code cleanup

cdouth99:


cdouth99:
just some small edits for a cleaner readme

TasoOneAsia:
Probably would be a good idea to include some references to the most requested docs page (ie menu, development).

cdouth99:
> Probably would be a good idea to include some references to the most requested docs page (ie menu, development).

Yeah I was just editing the looks, rather than contents. If you want, I'll add another version with updated information.

TasoOneAsia:
Might as well add those changes, if you could.

PR #491, opened 2021-11-08
feat(menu/actions): Add expiration to ban event data

antond15:
Add `duration` to data sent with `txAdmin:events:playerBanned` event.


antond15:
Changed `duration` to `expiration` because of permanent ban.
You can get duration by substracting current time and expiration.

tabarra:
You forgot the docs change :P  
Will do that in the next commit,  
Thanks for the PR!

PR #500, opened 2021-11-16
fix(docs/events): announcement documentation

Jaccosf:
The documentation for the announcement event was copied from the heal event and not modified. This will fix the documentation.


Jaccosf:
Wow, PR #500 🥇

TasoOneAsia (approved):

PR #507, opened 2021-11-20
feat(logger): adding full log customization

eduhdev12:
The user can trigger the event and mention anything they want without having executed: before the argument


TasoOneAsia (dismissed):
LGTM from an implementation standpoint, @tabarra any feedback for the potential use case?

tabarra:
Thanks for the PR.  
The locale stuff has been merged some time ago, and regarding the full custom log this is on hold.  
We need to plan on how to do that properly before.  
  
I'm closing this PR and will ship the log customization with the specification of the API.

PR #517, opened 2021-11-27
added(locale): Added ES locale. Note: already tested

carlos55ml:
Already tested. seems to work fine.

### DIFF
--- a/locale/es.json
+++ b/locale/es.json
@@ -37,250 +37,250 @@
     },
     "nui_menu": {
         "misc": {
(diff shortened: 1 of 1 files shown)


TasoOneAsia:
@carlos55ml Confirmed working and tested in-game?

carlos55ml:
Yes. Everything seems to display correctly.

tabarra:
All those 3 are too big:  
image

I'm still merging it, but if you don't mind try to find other phrases for the description in a way that it doesn't break to two lines.

PR #525, opened 2021-12-13
locale(es): changed line breaks sentences

carlos55ml:
Modified some lines so it does not break in two lines in menu.


tabarra:
Thanks for the PR.  
I think you forgot this big chungus here 
image

PR #528, opened 2021-12-14
Update hungarian translate

koko991:
kép_2021-12-14_161350
Reduce a few text lengths on request so as not to break the line.


tabarra:
Thanks!  
Can you also change this one please?
image

PR #534, opened 2021-12-20
feat(settings): added option to disable menu particle effects

AleksanderEvensen:
Didn't like the effects when entering no-clip, so I added the option to disable it, using the already existing Convar

### DIFF
--- a/docs/menu.md
+++ b/docs/menu.md
@@ -56,7 +56,7 @@ Convars configured in the settings page should not be set manually.
 **txAdmin-menuPtfxDisable**
 - Description: Determine whether to not play particles effects whenever an admin's player mode is changed.
 - Default: `false`
-- Usage: `+set txAdmin-menuPtfxDisable true`
+- Usage: `+setr txAdmin-menuPtfxDisable true`
 
 ## Commands
 **tx | txadmin**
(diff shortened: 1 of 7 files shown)


tabarra:
Sorry, but not enough people complained for me to add it to the settings page.

AleksanderEvensen:
Any particular reason why you shouldn't have the option to disable the particle effects. I'm just wondering. There is already a convar made for this purpose I just thought it would be nice to add the option to control this convar.

tabarra:
1. I want the settings page to be as simple as possible;
2. I don't want to deal with the config vault, it's a clusterfuck;
3. **As soon as I turn it into a setting, it's done and I can never touch it again for technical reasons;**
4. I like the effect, and it helps prevent admin abuse (turning on god mode during pvp);

AleksanderEvensen:
Ok, that's understandable. ill just keep using the convar in my own server.cfg, but could you look at the changes in the [scripts/menu/client/cl_player_mode.lua](https://github.com/tabarra/txAdmin/pull/534/commits/bce34f34a2d3ce3cefeabc1d24dcdeab9f72562e#diff-e7531473bd925f7feb353f11f56b4565e03ee26edd28346fae27ed709d0adb4f) file if the convar could remove the particle effects locally as well. Because now the convar will only remove particles for other players, and not for yourself.

tabarra:
This issue has been solved, you just need to update to v4.11

Issue #535, opened 2021-12-20
[Feature]: Set proxy mode true/false


ablax:
### Scope

Web

### Feature Description

Set proxy mode true/false

### Use Case

If someone want's to put txadmin behind proxy (cloudflare for example) they should still be able to get admins ip.

### Proposed Solution

Add Convar

Comet1903:
This is not possible, this is also how Tor works... Basically you can't get the admins IP... You can reverse proxy txAdmin on your own server, so you can view the admins IP in your proxy's log. (on my server I use nginx to do exactly what you want to do)
If that what I am saying here is not true then please correct me

ablax:
We are not talking about Tor here. When you run txadmin on your VPS, and an admin logs in, it will show the admin's IP in txadmin log. But having in mind that I do use nginx reverse proxy (just like you said) the txadmn no longer receives admin's IP, but instead it displays 127.0.0.1 (loopback address) because of the local proxy, and if you set it to add an header for forwarded-for address (just like cloudflare adds as well) and you set app.proxy to true koa automatically knows to use the header instead the IP making the request.

Rosw3lluk:
Im pretty certain if you use cloudflare, or i hhave in the past, the authentic origin ip is stored  as a header sent by cloudflare, which is the origin IP of the person making the request, and wont tell you cloudflares IP, PHPBB use this for there forum admin logs.

ablax:
Yes the IP is stored as a header from the proxy providers. And the purpose of this is to be able to get that IP, otherwise, txAdmin would be using Cloudflare's IP.

Doxylamin:
txAdmin should/could just use X-RealIP / X-Forwarded-For, if either of those are set.

ablax:
> txAdmin should/could just use X-RealIP / X-Forwarded-For, if either of those are set.

Pretty much it would be the same as enabling proxy mode. There is no difference.

Doxylamin:
Except there would be no need for such proxy mode and/or manual changes, as it would be detected automatically.

ablax:
If you enable proxy mode (you can just do it in the code without var, it's disabled on purpose) it will also detect it automatically and will work as you described it. That's the whole purpose of it

PR #536, opened 2021-12-20
Make proxy configurable

ablax:
Add convar to set proxy mode true/false

If someone want's to put txadmin behind proxy (cloudflare for example) they should still be able to get admins ip.

https://github.com/tabarra/txAdmin/issues/535


tabarra:
This is actually vulnerable to XFF IP Spoofing.  
https://github.com/expressjs/expressjs.com/issues/152#issuecomment-42468760  
https://security.stackexchange.com/questions/254199/how-to-prevent-spoofing-of-x-forwarded-for-header
  
This explains a bit better.

tabarra:
For reference: https://github.com/koajs/koa/issues/1413

ablax:
> This is actually vulnerable to XFF IP Spoofing. [expressjs/expressjs.com#152 (comment)](https://github.com/expressjs/expressjs.com/issues/152#issuecomment-42468760) https://security.stackexchange.com/questions/254199/how-to-prevent-spoofing-of-x-forwarded-for-header
> 
> This explains a bit better.

That's why it's configurable and disabled by default.

tabarra:
Optional or not, it's vulnerable to xff spoofing, and it's totally possible to solve that with the right settings.

ablax:
If you randomly enable proxy you may be vulnerable to xff spoofing, but if you have any kind of proxy behind your application and enable proxy and your txadmin is accessible only through the given proxy, you won't really be vulnerable to the exploit as the given proxy (either Cloudflare or just your reverse Nginx proxy) will be setting the header for you.

ablax:
> This is actually vulnerable to XFF IP Spoofing. [expressjs/expressjs.com#152 (comment)](https://github.com/expressjs/expressjs.com/issues/152#issuecomment-42468760) https://security.stackexchange.com/questions/254199/how-to-prevent-spoofing-of-x-forwarded-for-header
> 
> This explains a bit better.


```To prevent spoofing one need to be sure that

The web application accepts only connections from the reverse proxy, or at least only interprets x-forwarded-for when the request comes from the reverse proxy
The reverse proxy always removes any existing x-forwarded-for headers and always adds its own header
Web application and reverse proxy agree on what is a valid request and what is a valid header, i.e. things like HTTP request smuggling or similar attacks based on interpretation differences will not work. This might be done by properly normalizing the request in the reverse proxy and rejecting any wrong or unusual requests.
```


```The X-Forwarded-For (XFF) request header is mainly used for logging purposes as it enables the web server logs to show the original client IP address. The application could be modified to read this field and leverage it for access control though that would be a bad idea. Anyone on the internet could send a request directly to the web server (bypassing the proxy) and achieve website access by simply including an acceptable X-Forwarded-For request header. To further illustrate why this is undesirable, it would be like granting access based on the User-Agent string.
```
As stated in the stackexcahnge thread you provided. And in most cases if you have your txAdmin behind proxy, others won't be able to send direct requests, except if your setup just sucks, and you are revealing the endpoint to it.


Also, the GitHub link you provided is from 2014.

w33ts:
> > This is actually vulnerable to XFF IP Spoofing. [expressjs/expressjs.com#152 (comment)](https://github.com/expressjs/expressjs.com/issues/152#issuecomment-42468760) https://security.stackexchange.com/questions/254199/how-to-prevent-spoofing-of-x-forwarded-for-header
> > This explains a bit better.
> 
> ```
> 
> The web application accepts only connections from the reverse proxy, or at least only interprets x-forwarded-for when the request comes from the reverse proxy
> The reverse proxy always removes any existing x-forwarded-for headers and always adds its own header
> Web application and reverse proxy agree on what is a valid request and what is a valid header, i.e. things like HTTP request smuggling or similar attacks based on interpretation differences will not work. This might be done by properly normalizing the request in the reverse proxy and rejecting any wrong or unusual requests.
> ```
> 
> As stated in the stackexcahnge thread you provided. And in most cases if you have your txAdmin behind proxy, others won't be able to send direct requests, except if your setup just sucks, and you are revealing the endpoint to it.
> 
> Also, the GitHub link you provided is from 2014.

The problem is assuming people are being smart and secure about how they deploy the application. It's the FiveM community, don't expect a lot.

PR #560, opened 2022-01-12
locale(pl): Translate pending Polish phrases

HeXsss:
Added translations to untranslated parts of locale


tabarra:
Thanks for the PR.

PR #566, opened 2022-01-19
fix(helper): fix not being able to define multiple endpoints

AvarianKnight:
Not all endpoints _have_ to have the same endpoint, you can have however many endpoints you want for internal usage like defining port 8000 and only letting specific server access it for cache/proxy.

### DIFF
--- a/src/extras/helpers.js
+++ b/src/extras/helpers.js
@@ -198,9 +198,6 @@ function getFXServerPort(rawCfgFile) {
     if (!oneUDPEndpoint) throw new Error('You MUST have at least one <code>endpoint_add_udp</code> in your config');
 
     const firstPort = endpoints[0].port;
-    endpoints.forEach((m) => {
-        if (m.port !== firstPort) throw new Error('All <code>endpoint_add_*</code> MUST have the same port');
-    });
 
     //Check if the port is valid
     if (firstPort >= 40120 && firstPort <= 40150) {


tabarra:
So how would txAdmin know which one to use?

AvarianKnight:
> So how would txAdmin know which one to use?

Whichever endpoint is defined first is the one that is used (which is how it works currently)

So in the case of how we use it:

```bash
endpoint_add_tcp "0.0.0.0:30120"
endpoint_add_udp "0.0.0.0:30120"
endpoint_add_tcp "0.0.0.0:9800"
```

The connect endpoint would be `0.0.0.0:30120` while we have 9800 blocked from receiving any requests and is only used for proxy requests.

tabarra:
> Whichever endpoint is defined first is the one that is used

Sorry, but that sounds like hell to troubleshoot.  
A more robust solution is needed otherwise it wont be "Homer Simpson proof".
  
> (which is how it works currently)

Only because they all must match.

AvarianKnight:
> > (which is how it works currently)
> 
> Only because they all must match.

No, as it works fine with the check removed.

Out of context was this check added here because of the thought people could misconfigure or because someone actually did it?

It seems weird to block to default regular FXServer capabilities either way.

tabarra:
My intention was not to "block features", but to balance (dev time + troubleshooting time) against (features nobody use).  
You are overestimating even the average txAdmin user.  
This will only work fine if the first one the person declares is correct, but let's say they set the first one to `0.0.0.0:80` and the second one to `0.0.0.0:30120` (and no, we cannot trust the OS/ FXS will block using ports already in use), then what we will get is people complaining txadmin is crashing their server even though the server is totally working (as they can join).  
  
In txAdmin-land, everything stupid that _can_ happen, _will_ happen.  
Legit i had to replace all `localhost` for `127.0.0.1` as there were _multiple_ cases of people literally editing `/etc/hosts` to point their `localhost` to some external unreachable IP.  
  
Sorry, but just removing that check won't do.  
I may fix this limitation when I get to rewrite that entire frankstein-monster of a function

PR #576, opened 2022-02-03
Update de.json

JonasBade:
Fixxed a not correct German Translation


tabarra:
Thanks for the pull request.

PR #577, opened 2022-02-06
feat(core): initial support for custom launch scripts

kn0pee:
Hi,

I found a use case to have support to run a "script" such as a bat file before the server is launched. This could be used for things such as building a node project, or pulling a git repo.

This is by no means ready, however I wanted to collect some feedback, comments and concerns (especially regarding security) before I continued further.

My basic todo list is
- Better error handling, right now the server fails to launch if the pre-launch fails
- Other scripts such as post-launch and potentially custom scripts
- Permissions for configuring these settings / running custom scripts
- Being able to specify the directory script will run in (or defaulting to the base directory)

If you believe this is feasible, I am happy to flesh this out further. Feel free to reach out on here or discord if you would prefer.

### DIFF
--- a/src/components/configVault.js
+++ b/src/components/configVault.js
@@ -160,6 +160,7 @@ module.exports = class ConfigVault {
                 autostart: toDefault(cfg.fxRunner.autostart, null),
                 restartDelay: toDefault(cfg.fxRunner.restartDelay, null), //not in template
                 quiet: toDefault(cfg.fxRunner.quiet, null),
+                preLaunchScript: toDefault(cfg.fxRunner.preLaunchScript, null),
             };
 
             //Migrations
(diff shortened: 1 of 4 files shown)


tabarra:
Thanks for the PR.  
Sync external scripts are a big no no, also no error handling is being done.  
Need to detect exit codes, and fail to start the server if code !== 0.  
  
And although I see the value in the idea, I'm super worried about the server status handling, which tied with crash detection is extremely fragile and literally took me two years to make is work perfectly for all cases.  
Example of considerations:
- What status should be shown in the interface while it is running?
- What will happen if an manual restart is triggered while the script is running?
- Shouldn't this pre script be piped to the logger and live console?
- If it takes too long, what should we do?
- How do we even count the time? Would it be in the monitor module cooldown settings?

This is at least 10x more complicated than it looks.

kn0pee:
Hey Tabarra,

Thanks for the response. 100% agree that my implementation is not even close to something complete,I'll close it for now, as it is unlikely to progress, but it would be an awesome addition at some point!

If the external script issues are too complex (which you're right, it does seem that way) then potentially just direct git integration to pull a repository(s) before launching. Makes it a bit easier than needing to handle a completely external script.

Anyway as mentioned, thanks for your response. If you'd like to consider this in the future, I am happy to help either as a user or fleshing out the idea further.

tabarra:
Oh yes... git integration is coming!  
But we will also have to be careful with the git auto updater, because as above... i have no clue what's the best way of doing it.

Issue #594, opened 2022-03-02
[BUG] The discord bot will stop working April 30, 2022


tabarra:
**txAdmin/FXServer versions:**
All versions so far.
  
**Describe the bug**
txAdmin uses discord.js v11, which uses discord's v7 API that will be discontinued on April 30, 2022.
We need to migrate to something that uses the v10 API.  
  
This package was recommended, but no idea if it is any good: https://www.npmjs.com/package/eris

tabarra:
solved in v4.14.0

PR #596, opened 2022-03-04
fix(discord): v13 update

DamonOnYT:
This PR Addresses #594 where older versions of the Discord API will stop working. Currently, txAdmin runs Discord.JS v11 which is a far outdated version and uses old API's that will soon be discontinued.

Hopefully some time with some assistance from @tabarra I'd like to be able to add slash commands as a configurable option and work on some new commands. I don't want to go anywhere near adding settings without a bit of help haha

*webpack is a bitch, this tool me a lot longer than id like to admit*

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -17,10 +17,12 @@
         "@tabarra/discord": "^11.6.6",
         "axios": "^0.26.0",
         "boxen": "^5.1.2",
+        "bufferutil": "^4.0.6",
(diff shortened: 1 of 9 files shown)


tabarra:
Since I couldn't commit to your fork, I committed and merged locally: c2cb039b8ba715ba9e2d7208be85c3147d9024bd  
Thanks again for the help!

PR #610, opened 2022-04-03
Context Variable - Master Principal Identifier

ItsANoBrainer:
Add a new context variable users are able to utilize to automatically assign the master to any custom principal instead of just the default group.admin.


tabarra:
Where the docs at?  
What if admin has no ids?  
What if has multiple?

tabarra:
hello

ItsANoBrainer:
I've been away from development for a few weeks and getting back today, I'll update this.

tabarra:
Do you still need this?

ItsANoBrainer:
I am touching base with QBCore team to see if this is something they're still wanting. I will close this PR for now. Thanks

PR #612, opened 2022-04-04
feat(player): add actionRevoked event

cyrep:
version in `docs/events.md` needs to be changed since i couldn't find any documentation about versioning.


tabarra:
Thanks for the PR.  
I think it would be better to fire this event directly from the player controller so it's possible both to add some more information in the event, as well as prevent the event from firing if there was an error with the revoke action.  

What do you think?
https://github.com/tabarra/txAdmin/blob/master/src/components/playerController/index.js#L450

tabarra:
Will add this event in the next update, but the event data specs will be different.

PR #615, opened 2022-04-07
feat(spectate): set focus instead of teleporting admin to player

pachxl:
if a cheater is being spectated they can easily see the admin below them
by doing this the admin remains in their current position but still render all the ground around the spectated player, but the spectated player will have no idea they are being spectated

this is my first pr to anything so sorry if ive done anything wrong

### DIFF
--- a/scripts/menu/client/cl_spectate.lua
+++ b/scripts/menu/client/cl_spectate.lua
@@ -4,8 +4,6 @@
 if (GetConvar('txAdmin-menuEnabled', 'false') ~= 'true') then
     return
 end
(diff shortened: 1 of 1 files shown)


AquarDev:
Doesn't this break with how OneSync works?

pachxl:
> Doesn't this break with how OneSync works?

break in what way? @AquarDev

tabarra:
i think you wouldn't be able to hear people talking this way

pachxl:
> i think you wouldn't be able to hear people talking this way

will give this a test and let you know

PR #617, opened 2022-04-11
Improve passwords

MuDzYNN:
So basicly current you can use super insecure passwords like: password or something simmilar. Becuse there is only one validation - password length! So I decided to improve this a litte and first I only added more validation (one lowecase, one uppercase, one number and one special char). But later I think this not especialy good, so I added configuration for passwords to Settings tab. So now you can force using strong passwords to your server administrators and also don't need to use very complicated passwords for localhost purposes.

### DIFF
--- a/src/components/configVault.js
+++ b/src/components/configVault.js
@@ -117,6 +117,12 @@ module.exports = class ConfigVault {
                 menuEnabled: toDefault(cfg.global.menuEnabled, true),
                 menuAlignRight: toDefault(cfg.global.menuAlignRight, false),
                 menuPageKey: toDefault(cfg.global.menuPageKey, 'Tab'),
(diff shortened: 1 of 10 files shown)


roadtosixx:
Good PR

tabarra:
1. Making this configurable is kinda too much, we don't want to overwhelm the user with too many options;
2. If it was to be configurable, just a "entropy" or a checkbox to enforce more secure passwords, not configuring every single parameter;
3. We do have rate limiter in the auth webroutes, so bruteforcing is not as easy as it looks.  

With that said, I actually looked into adding password checks, but this needs to be done in 3 places and would make this "business logic" rule to be spread in 3 places.  
Just not worth the headache to do it now since soon we will migrate to React, which will make our life way way easier.

PR #630, opened 2022-05-11
Add ipv6 support to regex and allow bind-all for ipv6

Keyinator:
This will allow server owners to enable access for ipv4 and ipv6 by replacing
```
endpoint_add_tcp "0.0.0.0:30120"
endpoint_add_udp "0.0.0.0:30120"
```
with
```
endpoint_add_tcp "[::]:30120"
endpoint_add_udp "[::]:30120"
```
which is the ipv6 (and ipv4) equivalent of the above meaning it will bind to both specifications.


tabarra:
you literally haven't tested it.  
I'll do it myself.

PR #637, opened 2022-06-04
refactor(wrong name): Wrong name for locale

n1kkec:


tabarra:
Hmmm no? `si` stands for Sinhala.  
Please check https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes

n1kkec:
sl stands for Sri Lanka
si stands for Slovenia (slovenian domains are .si)

tabarra:
What you are talking about is domain name, which is not the same thing as language code.  
https://en.wikipedia.org/wiki/Country_code_top-level_domain

PR #640, opened 2022-06-07
Translated rest of da.json

Morfar22:
Translated rest of da.json


tabarra:
Thanks for the PR.  
You also edited some keys like `"options"` to `"muligheder"` which broke the translation, but i can fix it easily.

PR #641, opened 2022-06-07
tweak(client/cl_player_mode): make Ped/Player have no collision durin…

Mycroft-Studios:
…g noclip


tabarra:
I tested this locally and couldn't find a difference.  
Can someone else double test it?

TheFantomas:
> I tested this locally and couldn't find a difference. Can someone else double test it?

No I would say there is no difference. At least I didn't find any. I tested it on two people and we found no difference that was demonstrable or visible. On the other hand, maybe the 3 natives will help something and in the future this problem will not have to be solved.

PR #642, opened 2022-06-09
Godmode Changes

antmmatos:
For those who say godmode doesn't work


scripts/menu/client/cl_player_mode.lua
```diff
@@ -15,6 +15,14 @@ local function toggleGodMode(enabled)
         clearPersistentAlert('godModeEnabled')
     end
     SetEntityInvincible(PlayerPedId(), enabled)
```
tabarra, line 17: It should run `PlayerPedId()` only once and use it in the 9 commands.

PR #646, opened 2022-06-21
Update et.json

oradala:
"reject-temporary" BKeelustaja
"reject-permanent" Keelustaja

B <- shouldn't be in front of "Keelustaja"


tabarra (approved):

tabarra:
Thanks for the PR!

PR #650, opened 2022-07-01
Maybe works but nice update :D

Getpwnedbydolbajoba:
If something doesn't work then DM me (Lil Dolbajoba#4276), ty.


tabarra:
Doesn't work, package.json is not published with the built version.  
RIP

PR #651, opened 2022-07-06
Update fr.json

Mathu-lmn:
some translations were missing + corrected some french ones


tabarra:
Hey @Mathu-lmn thanks for the PR.  
Unfortunately this translation does not look good on the UI (example below), unless there is something wrong with the text, I recommend only changing stuff if it fits well with other options.  
Keep in mind the docs explain how can you test your translation before it gets merged into txAdmin.  
image

PR #660, opened 2022-08-11
refactor(client/cl_player_ids): Add event for toggleing player ids

justin-guenther:
I am running an rp-server and I'd like to know whoever is using nametags currently, so admins won't abuse it in some situations. The trigger will be called from another script to print a simple message into the chat for all admins.


tabarra:
Uhhh this is not how menu action logging works.  
I will add this feature manually and eventually, maybe in the v4.17 release, so will close this PR ok?

justin-guenther:
Sorry, didn't read the whole code correctly but you got the point :) Would be great having that feature in the v.4.17! Thanks!

Issue #613, opened 2022-04-05
[Feature]: Filter Started and Non-started resources


P4R4NORM4L:
### Scope

Web

### Feature Description

Have a button to display started or non started resources.

### Use Case

Some of us have a lot of resources, and when we restart a resource... some resources stop because of dependencies, so a quick way to bring them back online would be to have those buttons so we can toggle them back online. Also maybe not refresh the page once we click on start. 

### Proposed Solution

For the page refresh aspect of it, maybe using AJAX, but i'm not a web dev so I wouldn't know.

douth-c-archive:
Could you provide a example?

P4R4NORM4L:
yeah I can do that

P4R4NORM4L:
![image](https://user-images.githubusercontent.com/17122685/163253885-2d72778d-3d38-4171-b994-fe0c3bdb9b63.png)

P4R4NORM4L:
Hopefully this can help !

douth-c-archive:
thanks.

PR #665, opened 2022-08-26
feat(resource list): add stopped resource switch

CsokiHUN:
### I created the feature #613

### DIFF
--- a/web/main/resources.ejs
+++ b/web/main/resources.ejs
@@ -26,7 +26,7 @@
                     </label>
                     
                     <span class="form-text text-muted ml-3 mr-1">
-                        Stopped resources:
+                        Only stopped resources:
                     </span>
                     
                     <label class="res-switch c-switch c-switch-label c-switch-pill c-switch-success fix-pill-form">
--- a/web/main/resources.ejs
+++ b/web/main/resources.ejs
@@ -9,26 +9,26 @@
     <div class="col-md-7 mw-col6">
         <div class="card card-accent-info">
             <div class="card-header float-left">
-                <span style="font-size: large">All Resources</span>
+                <span class="h5">All Resources</span>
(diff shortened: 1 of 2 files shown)


web/main/resources.ejs
```diff
@@ -24,6 +24,15 @@
...
                         <span class="c-switch-slider" data-checked="On" data-unchecked="Off"></span>
                     </label>
+
+                    <span class="form-text text-muted ml-2" style="margin-top: calc(0.3rem + 2px)">
+                        Stopped resources:
+                    </span>
+                    &nbsp;&nbsp;
+                    <label class="c-switch c-switch-label c-switch-pill c-switch-success fix-pill-form">
+                        <input type="checkbox" id="stoppedResCheckbox" class="c-switch-input">
+                        <span class="c-switch-slider" data-checked="On" data-unchecked="Off"></span>
```
wowjeeez: Szia!
Reviewing on behalf of @tabarra. I think updating the labels here to `Show` and `Hide` would clear things up a bit. Otherwise I checked your code and I like the feature, I have a few code related ideas though.

wowjeeez (dismissed):
I think you should replace the callback functions to `() => ...` to make the code cleaner, since you don't rely on any context related info.

web/main/resources.ejs
wowjeeez: Maybe even renaming the switch to `Only stopped resources` or `Show only stopped` would be beneficial.

wowjeeez: Oh well

wowjeeez: Change the labels then

CsokiHUN: Hali!
I was a little tired last night, but I did it today, I hope it's good.

wowjeeez (approved):
Nice

tabarra:
Thanks for the PR @CsokiHUN, and thanks for the review @wowjeeez.

Issue #638, opened 2022-06-05
[FEATURE]: Staff icon in player list


napei:
### Scope

Web

### Feature Description

Provide a staff icon in front of the ID in the player list for the web portal to show staff members who are online.

### Proposed Solution

By placing the shield before the ID number, this can differentiate between players that have this emoji in their name as well to pretend to be staff.

![image](https://user-images.githubusercontent.com/8946502/172038144-6c3fa8f9-1a7a-4121-9239-a9b175f4c424.png)

This was with:
```css
.pping:before {
  content: '🛡️';
}
```

TasoOneAsia:
Implemented with v4.8.0

tabarra:
The issue is about the Web UI though, we only have it for the menu.

TasoOneAsia:
@tabarra whoops, want me to implement for Web UI real quick?

tabarra:
Meh this one we will have a hard time solving well without the React refactor, so for now probably better to leave it be.

tabarra:
Since I included this in #628, I'll close this as a duplicated.

PR #666, opened 2022-08-27
feat(player list): add admin badge

CsokiHUN:
### New feature by request #638 

- I think it's better that way after the name, it doesn't push the id

### DIFF
--- a/core/components/playerController/index.js
+++ b/core/components/playerController/index.js
@@ -591,6 +591,7 @@ export default class PlayerController {
                     name: p.name,
                     ping: p.ping,
                     identifiers: p.identifiers,
+                    isAdmin: p.isAdmin
                 };
             });
         } catch (error) {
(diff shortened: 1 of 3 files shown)


tabarra:
**Thanks for the PR @CsokiHUN.**  
Although i reckon this feature is needed, there is an issue with this PR:
The data hierarchy is inverted.  
txAdmin itself is the source of truth if some player is an admin or not, and in the proposed flow it's the txAdmin resource which is telling txAdmin weather a player is admin.  
This inversion is a somewhat dangerous pattern because if there are any vulnerabilities in the resource code, it would also vector it up to txAdmin itself.  
  
I'll close this PR because there is no right way of doing this right now, only wrong ways.  
In the hopefully very near future I will refactor the entire web playerlist (and player state management) which will allow us to do things the right way.  
  
Quick note for future self: players can have that emoji in their nickname, causing confusion, so on the new playerlist event remove that emoji from the nickname, as well as normalize and remove RTL, ansi escape and other weird control characters.

tabarra:
btw we are soon to refactor the entire web ui to be react-based, so I'm trying to maintain changes to a minimum

Lukeos11:
Maybe a good idea to add one for the master account

PR #672, opened 2022-09-04
added missing german translations

Mats4k:
Hey,
i added a few missing translations to stuff that was still in english.
The whole file should now be translated.

(Not using translator, native german)


tabarra:
Thank you for the PR.

PR #675, opened 2022-09-08
Added maxlength prop to confirmPassword input

oxince:
I've tried to enter a 60 characters long password and it was a bit buggy cause there was no maxLength to the "Confirm password" input so i've created this pull request.

### DIFF
--- a/web/parts/changePasswordModal.ejs
+++ b/web/parts/changePasswordModal.ejs
@@ -44,7 +44,7 @@
                         </label>
                         <div class="input-group">
                             <input type="password" autocomplete="new-password" class="form-control text-center"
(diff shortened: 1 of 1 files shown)


PC1up (approved):
I see your problem and also I dont see any possible conflicts with other PRs. But still @tabarra has to decide, just seems alright for me.

tabarra:
I think the field should not even have a max length...  
What if you paste a long password from your password generator and it just cuts the end silently without you realizing?!  
Preferably the field should allow the password but the javascript should warn you "wow this password is too long!"  
  
Now the question is, why did I add max length to all password fields?! No clue :(

PR #676, opened 2022-09-10
feat(menu): players blips

daZepelin:
Feature:
Menu option to disply players blips on map, coords fetched server side, so all players will be displayed no matter if onesync is on or off.

image
image

### DIFF
--- a/locale/hu.json
+++ b/locale/hu.json
@@ -160,9 +160,18 @@
             },
             "player_ids": {
                 "title": "ID mód váltása",
-                "label": "Toggle showing player IDs (and other info) above the head of all nearby players",
-                "alert_show": "Közeli játékosok ID-jének mutatása.",
-                "alert_hide": "Közeli játékosok ID-jének elrejtése."
+                "above_head": {
+                    "name": "Over Head IDs",
+                    "label": "Toggle showing player IDs (and other info) above the head of all nearby players",
+                    "alert_show": "Közeli játékosok ID-jének mutatása.",
+                    "alert_hide": "Közeli játékosok ID-jének elrejtése."
+                },
+                "map_blips": {
+                    "name": "Blips on map",
+                    "label": "Toggle players blips on map as well as their NetIDs.",
+                    "alert_show": "Showing players blips on map.",
+                    "alert_hide": "Hiding players blips on map."
+                }
             }
         },
         "page_players": {
--- a/resource/menu/client/cl_player_ids.lua
+++ b/resource/menu/client/cl_player_ids.lua
@@ -91,11 +91,11 @@ local function showPlayerIDs(enabled)
 
     isPlayerIDActive = enabled
     if not isPlayerIDActive then
-        sendSnackbarMessage('info', 'nui_menu.page_main.player_ids.alert_hide', true)
+        sendSnackbarMessage('info', 'nui_menu.page_main.player_ids.above_head.alert_hide', true)
         -- Remove all gamer tags and clear out active table
         cleanUpGamerTags()
     else
-        sendSnackbarMessage('info', 'nui_menu.page_main.player_ids.alert_show', true)
+        sendSnackbarMessage('info', 'nui_menu.page_main.player_ids.above_head.alert_show', true)
     end
 
     debugPrint('Show Player IDs Status: ' .. tostring(isPlayerIDActive))
@@ -116,6 +116,55 @@ RegisterNUICallback('togglePlayerIDs', function(_, cb)
 end)
 
 RegisterCommand('txAdmin:menu:togglePlayerIDs', togglePlayerIDsHandler)
+
+local playersBlips = {}
+---Refreshes blips on client
+---@param blipsData {[string]: {coords: {x:number,y:number,z:number,h:number}, name:string}}
+local function refreshPlayerBlips(blipsData)
+    for k, v in pairs(playersBlips) do
+        RemoveBlip(v)
+      end
+      playersBlips = {}
+      for playerId, v in pairs(blipsData) do
+        local blip = AddBlipForCoord(v.coords.x+0.01, v.coords.y+0.01, v.coords.z+0.01)
+        playersBlips[playerId] = blip
+        SetBlipShrink(blip, true)
+        SetBlipCategory(blip, 7)
+        SetBlipSprite(blip, 1)
+        SetBlipDisplay(blip, 4)
+        SetBlipScale(blip, 0.87)
+        SetBlipColour(blip, 2)
+        SetBlipFlashes(blip, false)
+        SetBlipRotation(blip, math.ceil(v.coords.h))
+        ShowNumberOnBlip(blip, playerId)
+        ShowHeadingIndicatorOnBlip(blip, true)
+        BeginTextCommandSetBlipName('STRING')
+        AddTextComponentString(v.name)
+        EndTextCommandSetBlipName(blip)
+      end
+end
+
+RegisterNetEvent('txAdmin:menu:refreshPlayerBlips', function(blipsData, enabled)
+    debugPrint('Received refreshPlayerBlips event')
+    refreshPlayerBlips(blipsData)
+    if enabled == nil then return end
+    if enabled then
+        sendSnackbarMessage('info', 'nui_menu.page_main.player_ids.map_blips.alert_show', true)
+    else
+        sendSnackbarMessage('info', 'nui_menu.page_main.player_ids.map_blips.alert_hide', true)
+    end
+end)
+
+local function togglePlayerMapBlipsHandler()
+    TriggerServerEvent('txAdmin:menu:showPlayerMapBlips', not isPlayerIDActive)
+end
+
+RegisterNUICallback('togglePlayerMapBlips', function(_, cb)
+    togglePlayerMapBlipsHandler()
+    cb({})
+end)
+
+RegisterCommand('txAdmin:menu:togglePlayerMapBlips', togglePlayerMapBlipsHandler)
 
 CreateThread(function()
     local sleep = 150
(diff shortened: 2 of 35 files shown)


locale/hu.json
```diff
@@ -160,9 +160,18 @@
...
+                    "label": "Toggle showing player IDs (and other info) above the head of all nearby players",
+                    "alert_show": "Közeli játékosok ID-jének mutatása.",
+                    "alert_hide": "Közeli játékosok ID-jének elrejtése."
+                },
+                "map_blips": {
+                    "name": "Blips on map",
+                    "label": "Toggle players blips on map as well as their NetIDs.",
+                    "alert_show": "Showing players blips on map.",
+                    "alert_hide": "Hiding players blips on map."
+                }
```
CsokiHUN, line 174:
```suggestion
                "name": "Fej feletti ID-k",
                "alert_hide": "Közeli játékosok ID-jének elrejtése."
                    "label": "Játékos ID (és egyéb információk) megjelenítésének kapcsolása az összes közeli játékos feje fölött",
                    "alert_show": "Közeli játékosok ID-jének mutatása.",
                    "alert_hide": "Közeli játékosok ID-jének elrejtése."
                },
                "map_blips": {
                    "name": "Blippek a térképen",
                    "label": "Játékosok blipjeinek, és NetID-nek kapcsolása a térképen.",
                    "alert_show": "Játékos blippek megjelenítve a térképen.",
                    "alert_hide": "Játékos blippek elrejetve a térképen."
                }
```
Hungarian translations added.

CsokiHUN (changes requested):
I will also check at the game script later.

resource/menu/client/cl_player_ids.lua
```diff
@@ -117,6 +117,55 @@ end)
...
+local playersBlips = {}
+---Refreshes blips on client
+---@param blipsData {[string]: {coords: {x:number,y:number,z:number,h:number}, name:string}}
+local function refreshPlayerBlips(blipsData)
+    for k, v in pairs(playersBlips) do
+        RemoveBlip(v)
+      end
+      playersBlips = {}
+      for playerId, v in pairs(blipsData) do
+        local blip = AddBlipForCoord(v.coords.x+0.01, v.coords.y+0.01, v.coords.z+0.01)
```
CsokiHUN, line 129: i think use [AddBlipForEntity](https://docs.fivem.net/natives/?_0x30822554) native because blip directly added to player ped. 
in the case of onesync, it may cause problems for remote players, the current solution may remain there.
_Write what you think._

resource/menu/server/sv_main_page.lua
```diff
@@ -189,6 +189,68 @@ RegisterNetEvent('txAdmin:menu:showPlayerIDs', function(enabled)
...
+    if type(playerBlipsData[tostring(src)]) == 'table' then
+      playerBlipsData[tostring(src)].blipsEnabled = not playerBlipsData[tostring(src)].blipsEnabled
+    else
+      playerBlipsData[tostring(src)] = {
+        blipsEnabled = true
+      }
+    end
+    TriggerEvent("txaLogger:menuEvent", src, "showPlayerMapBlips", allow, playerBlipsData[tostring(src)].blipsEnabled)
+    TriggerClientEvent('txAdmin:menu:refreshPlayerBlips', src, {}, playerBlipsData[tostring(src)].blipsEnabled)
+  end
```
CsokiHUN, line 211:
```suggestion
  local src = source
  local strSource = tostring(src)
  local allow = PlayerHasTxPermission(src, 'menu.viewids')
  if allow then
    if type(playerBlipsData[strSource]) == 'table' then
      playerBlipsData[strSource].blipsEnabled = not playerBlipsData[strSource].blipsEnabled
    else
      playerBlipsData[strSource] = {
        blipsEnabled = true
      }
    end
    TriggerEvent("txaLogger:menuEvent", src, "showPlayerMapBlips", allow, playerBlipsData[strSource].blipsEnabled)
    TriggerClientEvent('txAdmin:menu:refreshPlayerBlips', src, {}, playerBlipsData[strSource].blipsEnabled)
  end
  ```
  I think the code is better if there are not many tostrings 😄

File: resource/sv_logger.lua [hunk 4]
```diff
@@ -206,6 +206,13 @@ AddEventHandler('txaLogger:menuEvent', function(source, event, allowed, data)
         else
             message = "turned show player IDs off"
         end
+    elseif event == 'showPlayerMapBlips' then
+        if type(data) ~= 'boolean' then return end
+        if data then
+            message = "turned show player map blips on"
+        else
+            message = "turned show player map blips off"
+        end
```
CsokiHUN, line 215:
```suggestion
        message = "turned show player map blips " .. (data and "on" or "off")
```

CsokiHUN (changes requested):

CsokiHUN:
i checked game script, its good, but small changes required i think

SymonOS:
This would be really great if this came in TxAdmin. Is there any way I can test it?

resource/menu/client/cl_player_ids.lua
daZepelin: Yea, if onesync would be on, adding blips for each player ped entity (not coord) would result lots of blips on ped controlled by player (-1) instead of other players peds.

CsokiHUN: There won't be many blips, but you won't be able to retrieve peds from distant players.
```lua
local playerPed = GetPlayerPed(serverId)
if not DoesEntityExist(playerPed) then --ped invalid
--attach to coords
else
--attach to entity
end
```

daZepelin: This will always return true tho, as it if the player is far away it would be the same as GetPlayerPed(-1)
```lua
if not DoesEntityExist(playerPed) then --ped invalid
```

This would do tho, but I do not think it is neccessary really 🤔 
```lua
local playerPed = GetPlayerPed(serverId)
if playerPed  ~= PlayerPedId() then --ped invalid
--attach to coords
else
--attach to entity
end
```

CsokiHUN: I thought onesync doesn't work that badly 😢

tabarra:
@daZepelin Thanks for the PR, and for the effort making it really close to our usual design specs.  
But i'm not super sure how i feel about this.  
  
As a feature, it's cool, but the practicality and UX/UI don't feel super refined.  
One of our main concerns for most menu features is "how would that work if the server has over a thousand players online?", which is a real scenario for many txadmin servers.  
I'm worried about the server-side performance impact on the 2 seconds loop. Can you measure that and show us what it would looks like for 500 and 1000 players?  
On `sv_playerlist.lua` we don't even get the player coords, and we still doing dynamic delay to make sure we won't lag big servers.  
And for that playerlist (menu playerlist), we even gave up on doing coords on the server because of how much event data would be transferred, coords are super big and take many bytes to be sent through the events. Now consider 5 admins with  the player blips on, on 1000 players, how many mbps would that be uploaded?  
  
Also for 1000 players, what would the game map look like? is it visualizable? laggy?  
Maybe showing all individuals be too hard, maybe aggregate them into grids? maybe show a single one if you click on him in the menu playerlist?

daZepelin:
Thanks indeed for our concenrns and insights on this. I did some testing on this and it's performance as if there was 2000 players online and that's my conclusion

I did not really concern the network load on this, but 2000 players data would be around 80-90 kilobytes for one refresh to be sent to client if the values are sent as integer and nearly twice as much in float, but I think ints can do just fine. Sending data to client only for players who have their coordinates changed significantly would do even better, but that would cost server hardware load.

Talking of server load, I did not notice any noticable impact on it at all, as per resmon on server side, but always smaller yield per tick could used, that would obviously end up with higher timeout between refreshes, but that should be expected on servers with higher load.

Client performance with 2000 players did indeed get some impact as that dropped around 10 fps with over 2000 blips on the map. But I had a concept in my head for reworking the blips already and bundle players to a single blip if there are more than one of them within a specific radius. As what I and my team experienced in our server already, is that it's quite difficult to understand the spread of players around the map if the blips get on top of each other.

To be hornest I am pretty happy with this idea as staff from a few servers found it really useful, so I would be happy to work on it and improve it for it to be implemented in txadmin, hoping more servers would find use in it, but only if you @tabarra think it is worth a try and work.

tabarra:
Thanks for the tests.  
Right now my workload is more focused on backend core changes to prepare for a frontend migration and multiserver.  
As you can see in the open PR list, our menu stuff is kinda frozen since right now if there is any bug in the client side code it would be really really hard for us to even know it since it doesn't print on the console anymore.  
  
Before merging anything menu-related, we need to remove a few prints and convince Cfx to start printing txAdmin errors again, so a reasonable ETA for merging this PR would be 1~2 months, is that ok?

tabarra:
How about using this native to group/ungroup blips depending on player count in a specific grid area + zoom level?  
https://docs.fivem.net/natives/?_0x1363A998

Mycroft-Studios:
The bigger issue here would surely be the hard coded blip limit?

With there being a hard coded limit on the amount of blips, if a server already has a decent amount of blips (for example, vMenu alone adds a huge amount of blips), hitting that limit would be a quite a big concern at high player counts?

resource/menu/client/cl_player_ids.lua
NickReagan: Couldn't you use [`NetToPed(playerID)`](https://docs.fivem.net/natives/?_0xBDCD95FC216A8B3E)?

ItssJxstnDe:
I whould Like to See that Option❤️

tabarra:
Due to the complexity of this PR related to the usability of this feature with 1k+ players on the server, I decided to close it.  
Thank you very much for the PR though, really appreciate - even though I couldn't prioritize trying to solve the potential issues.  
For anyone in need of this, I think there are other alternative resources to do that in game or outside.

PR #678, opened 2022-09-11
feat(components/webserver): added support for proxification

MrSentex:
Proxy support added by setting the addrs in the `txAdminProxyAddresses` variable separated by `,`. No support for CIDRs as it would require a module and I leave that decision to tabarra.

The actual address of the client must be specified in the `x-txadmin-real-ip` header.

Example configuration:

```
set txAdminProxyAddresses "104.81.29.xxx, 92.10.23.xxx"
```

### DIFF
--- a/core/components/WebServer/ctxUtils.js
+++ b/core/components/WebServer/ctxUtils.js
@@ -235,6 +235,16 @@ export default async function WebCtxUtils(ctx, next) {
                 ctx.txVars.realIP = srcIP;
             }
         }
+    } else {
+        if (convars.proxyAddresses[ctx.ip]) { // Check if remote address is a proxy
+            const srcIP = ctx.headers['x-txadmin-real-ip'] || false;
+
+            if (srcIP) {
+                if (consts.regexValidIP.test(srcIP)) {
+                    ctx.txVars.realIP = srcIP;
+                }
+            }
+        }
     }
 
     //Functions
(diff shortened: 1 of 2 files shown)


CsokiHUN (approved):
is look a seems good

tabarra:
Hey @MrSentex thanks for the PR.  
Any particular reason why to use a custom txAdmin header instead of `x-forwarded-for`?

MrSentex:
> Hey @MrSentex thanks for the PR. Any particular reason why to use a custom txAdmin header instead of `x-forwarded-for`?

No real reason, I followed the pattern of the rest of the headers as `x-txadmin-identifiers`. `x-forwarded-for` is a better idea to follow the standard.

tabarra:
I think we better go with XFF, since it will be the zero-config alternative for most reverse proxies.  
But also i'm not sure about manually handling the ip/src validation.  
https://github.com/tabarra/txAdmin/blob/master/core/components/WebServer/index.js#L69

Koa itself has an option for that which can accept an array but must be configured correctly not to be vulnerable to XFF spoofing.

MrSentex:
The safest option I can think of would be to add a header for example: `x-txadmin-proxy-token` authenticating that the traffic is being proxied from an authorised server assuming that XFF will be valid. If it is configured incorrectly, the attacker will not have the key (If the key is not predictable. It could be generated automatically from the panel) and therefore will not be able to perform XFF spoofing. Configuring Koa for this seems unnecessary to me, we can work with `ctx.txVars.realIP` and it would be easier to make any modifications.

MrSentex:
But with the current configuration the user should not have to worry. We could raise an alert if the public IP address of the server is in the list of authorised IP addresses.

MrSentex:
The code should also be modified to follow the XFF standard as there may be more than one IP address concatenated.

PR #685, opened 2022-09-17
Create reverse_proxy.md

DesertFade:
Documentation about creating a reverse proxy for txAdmin to easily access it with a FQDN. This documentation is only meant for NGINX and commands + information is for linux.


S33G (approved):

spacevx:
You should use http 301 statut to redirect http requests to https

PR #686, opened 2022-09-21
Update tr.json

burakurer:
Turkish translation


tabarra:
Thanks for the PR.  
I just modified a few lines to prepare for the v5 update, if you don't mind, can you also translate those?  
https://github.com/tabarra/txAdmin/blob/develop/locale/tr.json  
  
edit: don't forget to target your PR to the `develop` branch :)

PR #696, opened 2022-10-28
fix(menu/players): fix scroll jump that could occur between player update cycles

TasoOneAsia:
Fixes intermittent players page scroll jump issue that could occur between player update cycles if a user was *not* on the first bucket. Now the player's page/grid will use the ***power of basic division*** to determine the bucket number to transition to upon an update player event.

_Note: This whole system could use a refactor, and the fakeLoading shenanigans and logic flow between Recoil state and component state is a mess that I don't want to jump into at the moment_.


tabarra (changes requested):
Need to review the reason why it is causing performance issues (in game) when using 500 fake players.  
The menu takes like a second to open.

tabarra (approved):

Issue #595, opened 2022-03-03
[Feature]: Option to suppress txAdmin announcements and utilize own system 


TasoOneAsia:
### Scope

Developer API

### Feature Description

Some servers would like to disable txAdmin notifications and instead utilize their own system bridged by the `txAdmin:events:announcement` event. At the moment, although you can hook into event, there is no way to purely disable txAdmin's standalone handling.

### Use Case

Server can utilize their own notification system instead of the txAdmin system.

### Proposed Solution

We can solve this by implementing a ConVar that would disable internal handling.

tabarra:
I would like this convar to be generic, where you would be able to specify more than one event handler to be bypassed.  
For example:
```bash
+set txAdmin-bypassEventHandler "playerWarned, announcement"
```  
  
Due to the complexity of the convar (csv is so hard isn't it) would be nice to have a custom inbox for it in the settings that would do some validating (`input.trim().split(/,\s*/)` and check individual parts).

tabarra:
Just adding back reference:  
> So i do not want the events to be bypassed, but this still brings the question, how do we handle people wanting to handle announcements/dm messages? I don't know, but that's much more a "ugly notistack bypass" than an event bypass per se.
Also, considering we only have announcements and DM to overwrite, this could totally be a checkbox in the settings.

araynimax:
@tabarra 
So how about implementing a convar that prevents anything notification related which can also be toggled via a checkbox in webinterface as you said?

Maybe something like:
`+set txAdmin-useCustomNotifications true`

araynimax:
Just to make sure it's clear: 
By Notification related i mean: announcements, dm's and player warns.

tabarra:
I think player warns should still be handled by txadmin, as it is (imho) the best implementation of a warn in fivem.  
And keep in mind that DM via snackbar is actually new in txadmin v5, not yet released.  
  
And i'm not so sure about joining both in a single setting, may be better to have two separate vars, and whatever it is done, should be restricted to changes on the server side only like preventing the events to being sent to the client or something like that.

araynimax:
I'm one of those developers who wants everything in one design, and I'm sure I'm not the only one. So there's nothing wrong with having the option to implement a custom warning imo.

araynimax:
> And i'm not so sure about joining both in a single setting, may be better to have two separate vars, and whatever it is done, should be restricted to changes on the server side only like preventing the events to being sent to the client or something like that.

Sounds right.
[...]

PR #698, opened 2022-11-01
feat(core): new Convar txAdmin-suppressEvents

araynimax:
Implemented the feature-request of #595
This also fixes a missing event trigger of
'txAdmin:events:announcement' from
announcements created in the game menu

### DIFF
--- a/resource/menu/shared.lua
+++ b/resource/menu/shared.lua
@@ -26,6 +26,20 @@ function txPrint(...)
   print(msg)
 end
 
+local supressedEvents = GetConvar('txAdmin-supressEvents', '')
+
+-- Used to check if a specific event should be suppressed
+function isSuppressedEvent(eventName)
+  for eventToBypass in string.gmatch(supressedEvents, '[^,%s]+') do
+      if string.lower(eventName) == string.lower(eventToBypass) then
+          return true
+      end
+  end
+
+  return false
+end
+
+
 CreateThread(function()
   debugModeEnabled = (GetConvar('txAdmin-menuDebug', 'false') == 'true')
 end)
--- a/resource/menu/shared.lua
+++ b/resource/menu/shared.lua
@@ -26,20 +26,6 @@ function txPrint(...)
   print(msg)
 end
 
-local supressedEvents = GetConvar('txAdmin-supressEvents', '')
-
--- Used to check if a specific event should be suppressed
-function isSuppressedEvent(eventName)
-  for eventToBypass in string.gmatch(supressedEvents, '[^,%s]+') do
-      if string.lower(eventName) == string.lower(eventToBypass) then
-          return true
-      end
-  end
-
-  return false
-end
-
-
 CreateThread(function()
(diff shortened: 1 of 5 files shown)


resource/menu/shared.lua
```diff
@@ -26,6 +26,20 @@ function txPrint(...)
   print(msg)
 end
 
+local supressedEvents = GetConvar('txAdmin-supressEvents', '')
```
TasoOneAsia: Small typo
```suggestion
local supressedEvents = GetConvar('txAdmin-suppressEvents', '')
```

resource/menu/shared.lua
```diff
@@ -26,6 +26,20 @@ function txPrint(...)
...
+-- Used to check if a specific event should be suppressed
+function isSuppressedEvent(eventName)
+  for eventToBypass in string.gmatch(supressedEvents, '[^,%s]+') do
+      if string.lower(eventName) == string.lower(eventToBypass) then
+          return true
+      end
+  end
+
+  return false
+end
```
TasoOneAsia: It is possible to slightly reduce time complexity by splitting and transforming the ConVar string into a map table. Then, once in the function, compare eventName against the string index in the map table. Reduces time complexity at the cost of space complexity.

If what I said makes no sense, don't worry about it, and keep it as is. This isn't a huge deal at all, and actual real-world use cases will see minimal difference. Change is up to you.

TasoOneAsia (commented):
Thanks for the PR! Left a couple preliminary comments before @tabarra checks it out. I'd also like to see some documentation for this new ConVar. As it stands, we don't have a central place that documents all of the available txAdmin convars. The menu convars are located in the `docs/menu.md` file, but it may be beneficial to consolidate all of the convars into its own `ConVar.md` file or something of that sort. I'll let @tabarra comment on that direction.

resource/menu/shared.lua
araynimax: Sounds good. I will adjust it that way.

resource/menu/shared.lua
araynimax: 👍

araynimax:
I added the convar documentation into the readme.md and docs/events.md. I think that are good places to find it as a dev.

tabarra:
Hey @araynimax, thanks for the contribution!  
  
From my post a few months ago I changed my mind about suppressing/bypassing events.  
If you look at the `feat/core-playerlist` you will see that right now I'm doing a more `txAdmin:event`-centric approach to tx > fxserver communications.  
I even removed a few commands in favor of events:  
image
  
So i do not want the events to be bypassed, but this still brings the question, how do we handle people wanting to handle announcements/dm messages? I don't know, but that's much more a "ugly notistack bypass" than an event bypass per se.  
Also, considering we only have announcements and DM to overwrite, this could totally be a checkbox in the settings.  
  
That being said, I'll close this PR now and we can discuss how to better handle those cases back in #595 .

PR #693, opened 2022-10-10
Update vietnamese translation.

nkyo:
create vi.json for vietnamese translation

PR #710, opened 2022-11-21
Update latest translation for Vietnameses

nkyo:
Updated latest translation for Vietnameses


tabarra:
Thanks for the pull request, but i noticed some keys are missing, and the label must be `Vietnamese` instead of `Tiếng Việt`.
image

nkyo:
Continue on PR #693

PR #726, opened 2022-12-07
Feature 725 - Add environment variable to skip printing IP banner

### LINKED ISSUE #725: [FEATURE]: Ability to hide the IP Banner dump
### Scope

Web

### Feature Description

When working on mods, and with the screen shared, the public IP address of the server is dumped on screen.

When I'm working in a dev environment, I don't want my private/personal IP address leaked.

### Use Case

Any time you start up the server, it dumps out a banner with public IP address information.

### Proposed Solution

An env var to disable banner dumping, maybe `BANNER_SKIP=1` (any value)

### Additional Info

Since a code is displayed in console when the txAdmin webui is loaded, this banner on startup doesn't hold much value except for first run. Subsequent runs can rely on various means of authenticating, including the code in console which is generally safe to display.

CodexHere:
Fixes #725

Adds a simple check for an env var presence to skip/bypass the IP banner dumping to console.

### DIFF
--- a/core/txAdmin.js
+++ b/core/txAdmin.js
@@ -3,7 +3,7 @@ import path from 'node:path';
 import slash from 'slash';
 
 import logger from '@core/extras/console';
-import { txEnv } from '@core/globalData';
+import { txEnv, convars } from '@core/globalData';
 
 import { printBanner } from '@core/extras/banner';
 import setupProfile from '@core/extras/setupProfile';
@@ -206,7 +206,7 @@ export default class TxAdmin {
         }
 
         //Once they all finish loading, the function below will print the banner
-        if (!process.env.SKIP_BANNER) {
+        if (!convars.skipIpBanner) {
             printBanner();
         }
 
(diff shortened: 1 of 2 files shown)


core/txAdmin.js
```diff
@@ -206,7 +206,9 @@ export default class TxAdmin {
         }
 
         //Once they all finish loading, the function below will print the banner
-        printBanner();
+        if (!process.env.SKIP_BANNER) {
```
TasoOneAsia: After some thought and discussion, it would probably make more sense to have this be derived from a convar launch arg to FXServer, rather than through an env var. Thereby matching current conventions in regards to configurable startup options. 

Like the following

https://github.com/tabarra/txAdmin/blob/e956d84b14e523e7c7165905d79b4743472c5b83/core/globalData.ts#L212-L226

tabarra:
I was thinking... perhaps a setting for "public-facing url" would be better.  
This could also be used by other stuff like the bot `/info` command have a link for the txadmin page for that player.  
Then if you just wanna hide your IP you can put in any gibberish that is mapped to your tx via DNS or whatever.

tabarra:
Just for the records, after 2 years, 2 months, 12 days, this is now an option through the `TXHOST_TXA_URL` env var.

PR #729, opened 2022-12-18
Convar for plateChange

TheFarFantomas:


tabarra:
After 7 months, do you still think this is required or the other resources already adapted?

ChatDisabled:
> other resources already adapted

Kinda is hard for third party resources to do this. Unless they listen to `txsv:logger:menuEvent` event, which too be fair no one is going to do. Considering txAdmin itself changes so it's only fair for the resource to change it back as well. Either a convar and or saving the previous plate and putting it back wouldn't be a bad idea.

tabarra:
With the help of @ChatDisabled (in DMs) I decided to just remove this plate-changing novelty feature.  
Therefore, this PR is not needed anymore.

PR #736, opened 2022-12-31
fix(resource/cl_logger): remove duplicate death casuse

PlouffeLuL:
Was looking for listeners and saw this so i figured might as well


tabarra:
Thank you :)

PR #744, opened 2023-01-24
Update de.json

PC1up:


Mats4k (approved):
This is a good translation, better then #745 in my opinion. Just here so you have another opinion on picking PRs.

tabarra:

PR #745, opened 2023-01-24
fix(locale/de.json): added german translation

SirLamaGott:
Added german translation ya


tabarra:

PR #746, opened 2023-01-24
fix(locale/nl.json): added nl translation

SirLamaGott:
ya


tabarra:
Thanks for the PR, but since it's "cross-contaminated" with the German changes I'll have to close it and cherry pick the dutch translation individually.

PR #749, opened 2023-01-25
🇫🇷 Vive La France DLC 🇫🇷

nooksmr:
This is the french locale of the txAdmin panel. After pull request, i'll delete my fork.

Have a pure day. 🫶🏼

Kept some english terms, as we use these terms, or for naming reasons on the panel


tabarra:
Thanks for the PR!

PR #755, opened 2023-01-27
Update de.json

MonsterZockerHD:
Super Jump translation added


tabarra:

PR #758, opened 2023-01-29
updated lt.json (Lithuanian)

Mindexas:
Updated some inaccuracies + updated translations of newly added functions/abilities.


tabarra:
thanks

PR #762, opened 2023-01-31
RedM support 

outsider31000:
just as a first commit  
* it adds a tp to way point using the menu  to support RedM without breaking other game modes
Ill be making more pr's just let me know if this is fine.

### DIFF
--- a/resource/menu/client/cl_base.lua
+++ b/resource/menu/client/cl_base.lua
@@ -13,15 +13,8 @@ lastTpCoords = false;
 RedM = false
 local isMenuEnabled = (GetConvar('txAdmin-menuEnabled', 'false') == 'true')
 
+IsGameRedM = GetGameName() == "redm"
 --Redm support
-CreateThread(function()
-  Wait(2000)
-  if (GetGameName() == "redm") then
-    RedM = true
-  else
-    RedM = false
-  end
-end)
 
 -- Check if menu is in debug mode
 CreateThread(function()
--- a/resource/menu/client/cl_base.lua
+++ b/resource/menu/client/cl_base.lua
@@ -10,10 +10,20 @@ isMenuDebug = false
 isMenuVisible = false
 menuPermissions = {}
 lastTpCoords = false;
+RedM = false
 local isMenuEnabled = (GetConvar('txAdmin-menuEnabled', 'false') == 'true')
 
+--Redm support
+CreateThread(function()
+  Wait(2000)
+  if (GetGameName() == "redm") then
+    RedM = true
+  else
+    RedM = false
+  end
+end)
 
--- Check if menu is in debug mode 
+-- Check if menu is in debug mode
 CreateThread(function()
   isMenuDebug = (GetConvar('txAdmin-menuDebug', 'false') == 'true')
 end)
@@ -31,7 +41,6 @@ local function checkMenuAccessible()
   return true
 end
 
-
 -- Register txAdmin command
 local function txadmin(_, args)
   if not checkMenuAccessible() then return end
@@ -45,6 +54,7 @@ local function txadmin(_, args)
     sendMenuMessage('openPlayerModal', targetPlayer)
   end
 end
+
 RegisterCommand('txadmin', txadmin)
 RegisterCommand('tx', txadmin)
 
@@ -70,7 +80,7 @@ TriggerServerEvent('txsv:checkAdminStatus')
 -- Triggered as callback of txsv:checkAdminStatus
 RegisterNetEvent('txcl:setAdmin', function(username, perms, rejectReason)
   if type(perms) == 'table' then
-    print("^2[AUTH] logged in as '"..username.."' with perms: " .. json.encode(perms or "nil"))
+    print("^2[AUTH] logged in as '" .. username .. "' with perms: " .. json.encode(perms or "nil"))
     menuIsAccessible = true
     menuPermissions = perms
     RegisterKeyMapping('txadmin', 'Menu: Open Main Page', 'keyboard', '')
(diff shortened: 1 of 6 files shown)


resource/menu/client/cl_base.lua
```diff
@@ -10,10 +10,20 @@ isMenuDebug = false
...
 
+--Redm support
+CreateThread(function()
+  Wait(2000)
+  if (GetGameName() == "redm") then
+    RedM = true
+  else
+    RedM = false
+  end
+end)
```
TasoOneAsia: Correct me if I'm wrong, as I cannot test this within RedM atm, but `GetGameName()` should be available at init script eval; any reason this thread needs to be here?

Also small nit, but it would be appreciated if the variable name was along the lines of `IsGameRedM`

```suggestion
IsGameRedM = GetGameName() == "redm"
local isMenuEnabled = (GetConvar('txAdmin-menuEnabled', 'false') == 'true')
```

resource/menu/client/cl_main_page.lua
```diff
@@ -63,42 +63,91 @@ if isMenuDebug then
...
+            debugPrint("^1Invalid  model requested: " .. model)
+            cb({ e = true })
         end
 
-        TriggerServerEvent('txAdmin:menu:spawnVehicle', model, modelType)
+        if IsModelAVehicle(model) then
+            ModelType = "vehicle"
+            GetVehicle()
+        else
+            ModelType = "ped" -- allow animals or peds
```
TasoOneAsia, line 121: Are any player "vehicles" considered ped/animals in RedM? If this is solely for arbitrary ped spawning, it may be better to just add a base menu option supporting that.

resource/menu/client/cl_main_page.lua
```diff
@@ -63,42 +63,91 @@ if isMenuDebug then
...
-            [15] = "heli",
-            [16] = "plane",
-            [21] = "train",
-        }
-        local modelType = types[VehicleType] or "automobile"
-        if model == GetHashKey("submersible") or model == GetHashKey("submersible2") then
-            modelType = "submarine"
+
+    if not RedM then -- rdr3 support
+        if isModelValid(model) then
```
TasoOneAsia, line 90: Typo

```suggestion
        if IsModelValid(model) then
```

resource/menu/client/cl_main_page.lua
```diff
@@ -63,42 +63,91 @@ if isMenuDebug then
...
+---if ped is in any vehicle then delete
+local GetVehicle = function()
+    local ped = PlayerPedId()
+    local oldVeh = GetVehiclePedIsIn(ped, false)
+    if oldVeh and oldVeh > 0 then
+        oldVehVelocity = GetEntityVelocity(oldVeh)
+        DeleteVehicle(oldVeh)
+    end
+
+end
```
TasoOneAsia: We don't use anonymous function variable assignment anywhere else in the menu scripts. To stick with current conventions and code style, prefer named local function definitions.

```suggestion
local function getAndDeleteVehicle()
    local ped = PlayerPedId()
    local oldVeh = GetVehiclePedIsIn(ped, false)
    if oldVeh and oldVeh > 0 then
        oldVehVelocity = GetEntityVelocity(oldVeh)
        DeleteVehicle(oldVeh)
    end
end
```

resource/menu/client/cl_main_page.lua
```diff
@@ -63,42 +63,91 @@ if isMenuDebug then
...
+        else
+            ModelType = "ped" -- allow animals or peds
+        end
+        TriggerServerEvent('txAdmin:menu:spawnVehicle', model, ModelType)
         cb({})
+
     end
 end)
 
+local LoadModel = function(model)
```
TasoOneAsia: See review comment above on anonymous function variable assignment

resource/menu/client/cl_main_page.lua
```diff
@@ -63,42 +63,91 @@ if isMenuDebug then
...
-            DeleteVehicle(oldVeh)
+    else
+        if not IsModelValid(model) then -- alow spawning horses or animals so we dont check if is a vehicle
+            debugPrint("^1Invalid  model requested: " .. model)
+            cb({ e = true })
         end
 
-        TriggerServerEvent('txAdmin:menu:spawnVehicle', model, modelType)
+        if IsModelAVehicle(model) then
+            ModelType = "vehicle"
```
TasoOneAsia, line 118: Unnecessary global def

TasoOneAsia (changes requested):
Thanks for the PR! I spotted a couple of items that need correction; Just a friendly reminder, prior to submitting future PRs, please ensure that you've tested behavior in-game. This PR was non-functional in-game due to a couple of typos.

Note: This was just a cursory review, and I'll try to provide more in-depth feedback when I get a chance.

Also, during my personal testing, I found an issue with handling invalid model names in FiveM. I don't have the time to debug this further, but I'll double-check that this is a regression.

resource/menu/client/cl_main_page.lua
TasoOneAsia: @tabarra You see value in making this another option?

resource/menu/client/cl_base.lua
outsider31000: there are many ways to do code , but sure  thats way better looking too.

resource/menu/client/cl_main_page.lua
outsider31000: vehicles can be  boats or wagons, peds horses animals, they all are created using the same native so i just named it ped
sure would be nice to have separated things in menu where users can know what can they spawn.

resource/menu/client/cl_main_page.lua
outsider31000: you right I should have follow the same way.

outsider31000:
in my latest pr It should be as it was suggested.

PR #765, opened 2023-02-09
Update ru.json - updating untranslated words in RU translation

egorlv:
This pull request aims to translate all untranslated words into Russian, ensuring that the meaning is accurately conveyed.


tabarra:
Thanks for the PR!

PR #775, opened 2023-02-27
Update tr.json

Raidr1212:
Translations for the new superjump player mode.
Edit: also added the whitelist messages.


tabarra:
Thanks for the PR!

PR #776, opened 2023-03-07
Added mouseover text to command history buttons

gegelendvay:
The mouseover text displays the whole command of each previously used command on hover, making it easier to identify longer-sliced commands and avoid mixing them up when trying to execute them on the console page.

Preview


tabarra:
Thank you.

PR #779, opened 2023-03-12
French Update

userzfr:
It is a update for the french langage becose the translater not good for all


tabarra:
Thanks for the PR!

PR #781, opened 2023-03-15
French Update

userzfr:
A new update following the update


tabarra:
Sorry, but it has been translated already.  
The new translation is available at the develop branch and you didn't notice because you committed to the master branch.

PR #789, opened 2023-03-29
Fixing noclip crashes

orel56000:
If you noclip for about 5 minutes full speed, it will crash your game, because the noclip loads a lot of interiors every frame and not releasing them. ending up with this crash: https://i.imgur.com/SaHSiqY.png
or the same with size 400.

This solution releases the interior once you get out of it.

### DIFF
--- a/resource/menu/vendor/freecam/camera.lua
+++ b/resource/menu/vendor/freecam/camera.lua
@@ -4,6 +4,7 @@ local SetCamRot = SetCamRot
 local IsCamActive = IsCamActive
 local SetCamCoord = SetCamCoord
 local LoadInterior = LoadInterior
+local UnpinInterior = UnpinInterior
 local SetFocusArea = SetFocusArea
 local LockMinimapAngle = LockMinimapAngle
 local GetInteriorAtCoords = GetInteriorAtCoords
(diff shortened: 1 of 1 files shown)


tabarra:
Ohh interesting.  
I'll get someone to test it, thanks!

tabarra:
@orel56000 After some investigation, i could not find a reason for the `LoadInterior()` to be used at all.  
If we remove that the noclip cpu time reduces in about 20~30%  
Do you mind validating if that words for you too?

orel56000:
@tabarra Actually, the first time I read the code I even asked myself why is it being used at all, and then I understood that it is because the ped is being teleported on intervals and the camera is moving on tick. So if you remove the LoadInterior usage, it will take about 0.5 seconds more to load the interior (if at all, because not all interiors are position dependent (I think. ??)) as it waits for the ped to teleport inside, which is actually not that critical, but I didn't want to make any _HUGE_ changes and just fix that one and keep the current view of how to noclip should work.

I guess both solutions are fine.

If it was up to me, I would have teleported the ped for every camera movement, because teleporting on interval makes more problems than it seems, like the proximity chat, etc..

tabarra:
Teleporting every 100 frames is a compromise.  
It reduces the resource CPU time A LOT.

tabarra:
Thanks again for the PR, I'll close this but reference it in the code in case we want to remove the `SetEntityCoords` in the future.  
For now I will just comment out the `LoadInterior` calls and this probably already fixes the crashes.  
  
For future reference:
I have tested the scenarios, and just moving camera (no `LoadInteriors` or `SetEntityCoords`) most interiors still load, except some garages (tunnels still load).  
With just `SetEntityCoords` everything appears to work, but without it we do need to `LoadInteriors`.

PR #790, opened 2023-04-01
Update sv.json

MasterVargen:
I fixed the whitelist message becuse they were in English.


tabarra:
Thanks for the PR.  
Unfortunately this file has already being translated, but you based your PR on the master branch so you probably missed it.  
Here's the link to the updated version:  
https://github.com/tabarra/txAdmin/blob/develop/locale/sv.json

PR #793, opened 2023-04-02
Spawn blimp crash

orel56000:
When spawning a blimp it creates it with server setter "plane" instead of "heli" This is caused because somehow `GetVehicleClassFromName()` for blimp return plane type instead of heli, even though if you check it in the network object viewer (on normal blimps created by `CreateVehicle()`) - it will say its a heli.

I see that the _solution_ for that issue for models such as `submersible` and `submersible2` is setting it specifically hardcoded, so I just added this model as well with the same solution.

### DIFF
--- a/resource/menu/client/cl_main_page.lua
+++ b/resource/menu/client/cl_main_page.lua
@@ -83,13 +83,20 @@ RegisterNUICallback('spawnVehicle', function(data, cb)
             [21] = "train",
         }
         local modelType = types[VehicleType] or "automobile"
-        if model == "submersible" or model == "submersible2" then
-            modelType = "submarine"
-        end
+      
+        local badTypes = {
+            ["submersible"] = "submarine",
+            ["submersible2"] = "submarine",
+            ["blimp"] = "heli",
+            ["blimp2"] = "heli",
+            ["blimp3"] = "heli"
+        }
 
-        if model == "blimp" or model == "blimp2" or model == "blimp3" then
-            modelType = "heli"
+        if badTypes[model] then
+            debugPrint("Model ".. model.." is included as a bad typed model, setting it from "..modelType.. " to "..badTypes[model])
+            modelType = badTypes[model]
         end
+    
             
         -- collect the old velocity
         local ped = PlayerPedId()
--- a/resource/menu/client/cl_main_page.lua
+++ b/resource/menu/client/cl_main_page.lua
@@ -83,12 +83,12 @@ RegisterNUICallback('spawnVehicle', function(data, cb)
             [21] = "train",
         }
         local modelType = types[VehicleType] or "automobile"
-        if model == GetHashKey("submersible") or model == GetHashKey("submersible2") then
+        if model == "submersible" or model == "submersible2" then
             modelType = "submarine"
         end
-            
-        if model == GetHashKey("blimp") then
-            modelType = "heli"        
+
+        if model == "blimp" or model == "blimp2" or model == "blimp3" then
+            modelType = "heli"
         end
             
         -- collect the old velocity
--- a/resource/menu/client/cl_main_page.lua
+++ b/resource/menu/client/cl_main_page.lua
@@ -86,6 +86,11 @@ RegisterNUICallback('spawnVehicle', function(data, cb)
         if model == GetHashKey("submersible") or model == GetHashKey("submersible2") then
             modelType = "submarine"
         end
+            
+        if model == GetHashKey("blimp") then
+            modelType = "heli"        
+        end
+            
         -- collect the old velocity
         local ped = PlayerPedId()
         local oldVeh = GetVehiclePedIsIn(ped, false)


orel56000:
I actually have better idea in mind:
I've made this list of model types, which is more accurate than `GetVehicleClassFromName()` in this use-case:

https://pastebin.com/69F14s3M

which can be used (server sided) to determinate what is the model type of the model instead of passing it through the event.

tabarra:
Thanks for the PR!  
My game still crashes even when setting the modelType to `heli` 😢 
What could it be?  
  
Also there are `blimp2` and `blimp3` as well.  
Perhaps would be better to make a "overwrites table" adding the submersibles instead of doing many IFs.  
For context, i'm using `sv_enforceGameBuild mptuner`.

Mycroft-Studios:
It crashing at both could be an issue with the native?

Mycroft-Studios:
It crashing with both could be an issue with the native maybe? Since the docs define the types as:
automobile, bike, boat, heli, plane, submarine, trailer, and (potentially), train.

orel56000:
> Thanks for the PR! My game still crashes even when setting the modelType to `heli` 😢 What could it be?

Hmm, that not making any sense to me because I checked (with runcode) it and it worked for me. 
When I spawned a blimp with plane it crashed, and when I spawned with heli it didn't... Maybe I did something wrong with the pr but it shouldn't crash, I'll check again later and let you know

orel56000:
`CreateVehicleServerSetter("blimp", "heli", GetEntityCoords(GetPlayerPed(3)), 0.0)` - Works
`CreateVehicleServerSetter("blimp", "plane", GetEntityCoords(GetPlayerPed(3)), 0.0)` - Crash

Even if my code is integrated with txadmin, when activating menuDebug it still sends it like a plane
image

orel56000:
Okay, I managed to get it working!
I just copied the code from the lines above, and I thought that it is correct.
So the problem is this if statement:
` if model == GetHashKey("blimp") then`

The problem is that `local model = data.model` is not hash, but a string!

When I change my lines to `if model == "blimp" then` it started working.
So I guess this quick fix is not working either:
`if model == GetHashKey("submersible") or model == GetHashKey("submersible2") then
            modelType = "submarine"
end`

tabarra:
Nice job.  
Do you mind refactoring it to be a lookup table?  
As in, first do the `types[VehicleType]` thing, and then you replace the submersible and blimp IFs for a lookup table containing the model name and a vehicle type to force.

orel56000:
> Nice job. Do you mind refactoring it to be a lookup table? As in, first do the `types[VehicleType]` thing, and then you replace the submersible and blimp IFs for a lookup table containing the model name and a vehicle type to force.

I have a better solution, if you would like to, instead of using the current solution which is depending on `GetVehicleClassFromName()`, I've created a list of the correct model types of all the models (up to gamebuild 2699), depending on `GetVehicleType()`, which is a more suitable to the current problem.
And if a player tries to spawn a vehicle that is not from the list, then it would use the old method.

This is the list:
https://pastebin.com/69F14s3M

tabarra:
I just talked to a FiveM dev that clarified that vehicle class is a totally different thing to vehicle type, and `CreateVehicleServerSetter` requires vehicle type instead of class.  
I do not want to use a custom table as maintaining that seems like a nightmare.  
I'm waiting their reply to see what is the correct way to approach this.  
FYI @Mycroft-Studios

Mycroft-Studios:
> I just talked to a FiveM dev that clarified that vehicle class is a totally different thing to vehicle type, and `CreateVehicleServerSetter` requires vehicle type instead of class. I do not want to use a custom table as maintaining that seems like a nightmare. I'm waiting their reply to see what is the correct way to approach this. FYI @Mycroft-Studios

The reason Class was used, is because CFX just kinda forgot to add a native for getting said Type, the only one that exists is getting the type of an already spawned vehicle (`GetVehicleType`), which is useless. so while they added `CreateVehicleServerSetter` to the server, there is nothing for said type on the server, only client, and even then, class was the best way for it

Mycroft-Studios:
After a quick check, there is literally no native at all, for getting this exact type. Converting Model to Class to Type is literally the only non-hardcoded way it seems

orel56000:
Having a server-sided native to get the vehicle type BEFORE the vehicle is spawned is not making any sense at all. The server can't know the vehicle type, because it is not _containing_ the `vehicles.meta` files. This is the whole reason for the native `CreateVehicleServerSetter()`, because they can't know it, they tell you to input a vehicle type based on your information.

If there was a way to know the vehicle type in the server side, then they would have just change the `CreateVehicle()` native from RPC to ServerSetter, without all the troubles of inputting a type.

tabarra:
@orel56000 that does not need to happen on the server side, it's actually the client which resolves the class (to be changed to type) before sending it to the server.

Kippstuhl:
123

PR #795, opened 2023-04-04
hu.json whitelist translate

KacsaTvT:
Hi guys, there is a new (and maybe the first) translation for the new whitelist lines.

### DIFF
--- a/locale/hu.json
+++ b/locale/hu.json
@@ -29,29 +29,29 @@
     },
     "whitelist_messages": {
         "admin_only": {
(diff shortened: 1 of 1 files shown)


tabarra:
Aw not quite the first, but thanks for the PR :P  
This has already being translated, but you were looking at the master branch instead of the develop one.  
Here is the link of the updated file, which you can also submit a PR to change if it is not quite right.  
https://github.com/tabarra/txAdmin/blob/develop/locale/hu.json

KacsaTvT:
oh, i didn't see it. then never mind :D

PR #798, opened 2023-04-09
wip(resource): NoClip RedM prompts | Native naming | Expected values

ChatDisabled:
- Changed `CTRL_SLOW = 0x8AAA0AD4` to `0x580C4473` since the previous hash was middle mouse button somehow
- Updated some very old natives to their new naming scheme
- Utilise rdr prompts to indicate movement controls (FiveM _equivalent_ for Instructional buttons)
- Updated some native so they get their expected values (boolean instead of an integer and vice versa, also the `RegisterCommand` restricted value.)

Sorry didn't know you were also actively developing for it 🙃. Will break PRs more down from now on and keep it structured.


tabarra:
Thanks ❤️

PR #800, opened 2023-04-12
fix(client): super jump

ChatDisabled:
Still kind of confused on the stamina bug part since this 100% works on both FiveM and RedM. Since all natives used needs to be ran on each frame there is no need to disable them. This means we could also put `clearPersistentAlert` behind the while loop so it automatically gets ran when it gets disabled.

`SetPedMoveRateOverride` only gets set at 1.75 max in decompiled scripts in both FiveM and RedM. Unsure where the 10.0 comes from. Definitely works in FiveM though but seems to be some kind of placebo affect in RedM? Otherwise might need to test with `Citizen.Invoke`.

Get the player ped every frame in the odd chance the ped changes because of death etc👍🏼.

### DIFF
--- a/docs/dev_notes.md
+++ b/docs/dev_notes.md
@@ -48,7 +48,7 @@
         - [x] scaleform/prompt
     - [x] god mode
     - [x] super jump
-        - [ ] fix stamina bug
+        - [x] fix stamina bug
     - [x] normal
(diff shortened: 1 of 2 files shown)


tabarra:
Thanks ❤️

PR #806, opened 2023-04-24
Update CS.json to translate another English words

Mxthesss:
To Czech language ;)


tabarra:
Thanks for the PR, but you edited the outdated master file.  
If you still feel like it needs editing, please edit the updated file below:  
https://github.com/tabarra/txAdmin/blob/develop/locale/cs.json

PR #809, opened 2023-05-03
added Gold cores RedM "BOOST"

outsider31000:
using BOOST from the menu if RedM will check if player is on a horse or on foot and apply Gold cores


tabarra:
Thanks for the PR.  
I ended up changing the behavior a bit, and the code quite a bit. So i'll just merge manually.

PR #811, opened 2023-05-06
Update zh.json

Cata-a1138:
Update translation


tabarra:
Thank you for the PR.

PR #813, opened 2023-05-10
Add txAdmin Mongolian Locale Option

Bat-EnkhWalter:
image
image
image


tabarra:
Thanks for the pull request, but I cannot merge it for two reasons:
- It has "roleplay talk" like "tsunami" instead of "restart";
- the `restarter.schedule_warn` and `restarter.schedule_warn_discord` are not in the correct format. What comes before `||||` is singular (1 minute remaining) and what comes after is plural (more than a minute).

I'll leave this PR open in case you want to edit it.

Bat-EnkhWalter:
@tabarra the issues have been fixed

tabarra:
Since the scheduled warn message was still invalid, I reverted it back to English because I'm in hurry to release txAdmin v6.  
But feel free to make a new pull request to change that if you feel like it.

PR #818, opened 2023-05-16
refactor(scripting): avoid redundant code

spacevx:
Avoid redundant code and condition branching


tabarra:
Thanks for the PR.  
I'll review after v6 is released, don't want to risk breaking stuff right now.

tabarra:
Thanks for the PR.  
I didn't merge it because I wanted to change basically everything, and one of the changes I did not like, so I made a commit myself.

PR #821, opened 2023-05-19
Server side command logger

TehSteel:
Server side command logger


tabarra:
Thanks for the PR.  
IDK if i want this interface to continue at all, we should migrate to exports soon enough, so i'll keep this open while i think about it.

TehSteel:
First of all thanks for responding, I'll watch this PR in case you will replay.

PR #843, opened 2023-06-28
Completely update lithuanian locale (lt.json)

R0JEA:
Updated remaining translations in Lithuanian language


tabarra:
Thanks for the PR.

PR #846, opened 2023-07-07
fix(core): shutdownNoticeDelay not having correct value

Mycroft-Studios:
the previous `??` was setting the variable to `null` causing servers to instantly restart. heavily breaking some things


Mycroft-Studios:
Closed due to being fixed in https://github.com/tabarra/txAdmin/commit/f75bd9683fd20abd68ad767e28ab4e603684a432

PR #851, opened 2023-07-16
fix(nui/PlayerModel): require OneSync for bring and goto

Mycroft-Studios:
GetPlayerPed is only available on the server-side, if OneSync is enable - note that it works on both legacy and infinity. having onesync off, which result in an invalid entity, resulting in the target coords being 0,0,0.

note: untested, but should work just fine


locale/en.json
```diff
@@ -77,7 +77,8 @@
             "not_enabled": "The txAdmin Menu is not enabled! You can enable it in the txAdmin settings page.",
             "announcement_title": "Server Announcement by %{author}:",
             "directmessage_title": "DM from admin %{author}:",
-            "dialog_empty_input": "You cannot have an empty input."
+            "dialog_empty_input": "You cannot have an empty input.",
+            "onesync_error": "This option requires OneSync to be enabled."
```
tabarra, line 81: Note to future self: perhaps remove `nui_menu.page_main.vehicle.onesync_error`

tabarra (changes requested):
No need to do anything, I'll take care of it when merging.  
Just marking as change pending to make sure I see it.

PR #856, opened 2023-07-28
Add support to Windows Server 2022

Red-Killer:


tabarra:
This file is straight from @sindresorhus/windows-release, but with async functions.  
So instead of changing stuff in it, what needs to be done is re-sync with his version and change it to support async calls.

PR #868, opened 2023-08-22
Add japanese translation

Yagami-Huki:
txAdmin1
txAdmin2
txAdmin3


tabarra:
Thank you very much for the PR.  
The file passed my automated checking, so it is probably good to go, but I'll review when I'm closer to releasing the next update.

Yagami-Huki:
I found a small mistake in the translation and fixed it.
I apologize for not checking it properly.

PR #869, opened 2023-09-04
Fix for 0 hour ban expiration display bug

gegelendvay:
Players have been asking about a discrepancy where the ban duration displayed as 0 hours. To solve this, I've added a minute unit. The issue came up when the ban was set to be revoked within an hour.

Before:
image

After:
image

### DIFF
--- a/core/webroutes/player/checkJoin.ts
+++ b/core/webroutes/player/checkJoin.ts
@@ -191,6 +191,7 @@ function checkBan(
         if (ban.expiration) {
             const humanizeOptions = {
                 language,
+                largest: 2,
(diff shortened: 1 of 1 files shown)


tabarra:
Thanks for the PR.  
Should we also add `largest: 2` to not show too many numbers?

gegelendvay:
I got these sample outputs with the use of the `largest: 2` param from a small test script I made:
image
Without the parameter, it would look like this:
image

I am fine with both, so the decision is up to you.

tabarra:
Thanks again for the PR!

Issue #421, opened 2021-08-21
UI Should Scale for higher resolution monitors 


TasoOneAsia:
The menu was designed primarily using a 1080p monitor and works well for that resolution. At 1440p, elements can start to feel a bit small. At 4k, items are absolutely tiny and need to be scaled up for the user.

Solutions:
* Apply the non-standard `zoom` CSS property available in CEF. (This is a non-standard feature and can be subject to change in the future, would prefer to not use if possible. Unfortunately, this will also probably be the easiest to implement seamlessly with minimum changes.)
* Adjust the breakpoint system and to be more suited towards desktop development and not mobile responsiveness. (Adjusting breakpoints like this often leads to small CSS issues that can become aggravating to fix).

*Note: Often when developing for regular browsers, users with 4k screen will enable a browser setting to enable auto scaling. Since this is CEF, there aren't many good solutions for enabling these scaling options programmatically*

TasoOneAsia:
Although zoom and transform property works in general, popover elements are not handled properly. This is fixed in #439 as it utilizes Popper v2. When that is merged, this should be fairly easy to implement well.

tabarra:
We also need to do a `max-width` on the entire thing since some people have ultrawide screens.

PR #870, opened 2023-09-06
Rescale UI Elements for 4K Displays

frvgs:
I'm used to playing on 4k screens and the Menu UI was almost unreadable

as described in this [issue](https://github.com/tabarra/txAdmin/issues/421) I used media queries targeting displays with a min-height of 2160px to rescale the Menu

Tested UI elements on both standard and 4K resolutions and verified that UI adjustments are applied only on displays with a height of 2160px or greater.

**before**
before

**after**
after

### DIFF
--- a/nui/src/components/MainPage/MainPageList.tsx
+++ b/nui/src/components/MainPage/MainPageList.tsx
@@ -46,6 +46,9 @@ const BoxFadeTop = styled(Box)(({ theme }) => ({
   position: "relative",
   bottom: listHeight + fadeHeight - 4,
   height: fadeHeight,
+  '@media (min-height: 2160px)': {
+    bottom: 600 + fadeHeight - 4,
+  }
 }));
 
 const BoxFadeBottom = styled(Box)(({ theme }) => ({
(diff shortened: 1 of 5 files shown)


tabarra:
That's awesome, thanks.  
@TasoOneAsia mind reviewing it?

iSentrie:
image

frvgs:
mb, those weren't the right screenshots, here's what it looks like before and after the rescale on 4k displays : 

**before**
before
**after rescale**
after2

iSentrie:
a bit weird that font size differs 
image
should be
image

frvgs:
> a bit weird that font size differs image should be image

Thank you for your feedback, it's fixed

textfix

tabarra:
Hey @frvgs, thanks for the PR.  
With txAdmin v7 and new UI, there is a chance that soon I'll migrate the ingame UI to use the same library and visual styles of the web UI, and with that, everything will be measured in REM units, solving this issue properly.  
  
As of now, I'm not sure if this will be done, or when this will be done, so I'll leave this PR open in case I decide to merge it instead.

Z3rio:
> Hey @frvgs, thanks for the PR. With txAdmin v7 and new UI, there is a chance that soon I'll migrate the ingame UI to use the same library and visual styles of the web UI, and with that, everything will be measured in REM units, solving this issue properly.
> 
> As of now, I'm not sure if this will be done, or when this will be done, so I'll leave this PR open in case I decide to merge it instead.

Yep ^^ This should be one of the better solutions.
To turn all `px` -> `rem`, and then just update the rem unit at the `head` object.
Way less hacky than this, and more uniform!

I could perhaps help out with this if I have some time over

PR #872, opened 2023-09-07
fix: dynamically create prompt title 

ChatDisabled:
For some reason whenever an UI element showed text (in this case the settings menu), it would overwrite the variable `redmPromptTitle`. Don't know how this hasn't been picked up by anyone else or maybe this is a problem on my end? Since I created the code myself this might as well simply be my lack of understanding core RDR behaviour (development wise).

I've added the `txcl:setPlayerMode` event to listen if noclip is being fired and recreate the title string when so. Theoretically speaking no performance downside. Same goes for spectate. Those are the only 2 functionalities using prompt titles.

<details><summary>Example images of the error</summary>
<p>

image
image

</p>
</details>

After the PR it should go back to 
image

### DIFF
--- a/resource/menu/client/cl_spectate.lua
+++ b/resource/menu/client/cl_spectate.lua
@@ -185,10 +185,12 @@ local keysTable = {
 }
 
 local redmInstructionGroup, redmPromptTitle
-if IS_REDM then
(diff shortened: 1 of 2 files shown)


ChatDisabled:
Moved the code to the corresponding functions.

PR #886, opened 2023-12-03
feat(menu): add keymapping for tp to waypoint

btwlouis:
This pull request introduces a Key Mapping feature for teleporting to waypoints. Users can now easily assign keys to the teleportation process.

not tested, but should work fine

### DIFF
--- a/resource/menu/client/cl_main_page.lua
+++ b/resource/menu/client/cl_main_page.lua
@@ -250,3 +250,8 @@ RegisterNetEvent('txcl:clearArea', function(radius)
     -- after found nativedb info. Maybe needs research lmao?
     ClearAreaLeaveVehicleHealth(curCoords.x, curCoords.y, curCoords.z, radiusToFloat, false, false, false, false, false)
 end)
+
+-- Teleport to the current waypoint via command for key mapping
+RegisterCommand('txAdmin:menu:tpToWaypoint', function()
+    TriggerServerEvent('txsv:req:tpToWaypoint')
+end)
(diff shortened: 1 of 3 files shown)


tabarra:
Now i'm wondering... why haven't I added optional keymappings to most menu features?!  
I'm thinking there was probably a reason, like Cfx saying that there was a limit or something because it's a default resource?!  
  
🤔

resource/menu/client/cl_main_page.lua
```diff
@@ -250,3 +250,8 @@ RegisterNetEvent('txcl:clearArea', function(radius)
     -- after found nativedb info. Maybe needs research lmao?
     ClearAreaLeaveVehicleHealth(curCoords.x, curCoords.y, curCoords.z, radiusToFloat, false, false, false, false, false)
 end)
+
+-- Teleport to the current waypoint via command for key mapping
+RegisterCommand('txAdmin:menu:tpToWaypoint', function()
```
KaseyFitton, line 255: this should also probably be added the list of commands that get their suggestions removed.

Mycroft-Studios, line 255: this should probably be added to the list of commands that get their suggestions removed..

btwlouis:
> Now i'm wondering... why haven't I added optional keymappings to most menu features?! I'm thinking there was probably a reason, like Cfx saying that there was a limit or something because it's a default resource?!
> 
> 🤔

So I don't know that there is a limit, esx uses more than 3 in its latest version and it runs without any problems.

Mycroft-Studios:
> So I don't know that there is a limit, esx uses more than 3

ESX isnt a default resource. by "default" tabby is referring to resources that are bundled with the fxserver and/or cfx made (such as chat)

resource/menu/client/cl_main_page.lua
btwlouis: Fixed in the last commit.

tabarra:
Huh... apparently my local squash merge didn't detect this as merged.  
Either ways, thanks for the PR!

PR #895, opened 2024-01-16
fix(resource): check if resource description breaks json.encode and remove checking a specified dev

itsmaty:
Discussed with @tabarra on discord here:
https://discord.com/channels/577993482761928734/1140025605740564581/1196835709978411039

Basicly i removed the checking for just one developer that seems to have that broken resource description and instead add a general checking if the resource description breaks json.encode.

@tabarra said that the issue with the broken unicode was probably fixxed on cfx side and it may not be a problem at all anymore that resource description breaks it. But decided to do a pr anyways.

### DIFF
--- a/resource/sv_main.lua
+++ b/resource/sv_main.lua
@@ -123,11 +123,22 @@ local function txaReportResources(source, args)
     for i = 0, max do
         local resName = GetResourceByFindIndex(i)
 
-        -- Hacky patch added because a particular resource from this developer had a broken 
-        -- unicode in the resource description, which caused json.encode to fail.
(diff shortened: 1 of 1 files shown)


tabarra:
Thanks so much for the research and PR, but I have tested and indeed the issue has been solved, so that hacky patch is not needed, neither your PR.  
  
Test `fxmanifest.lua`:
```lua
author 'ok\xE2\x96\x93'
version 'broken\xE2\x96'
description 'broken2\xE2'
```  
  
txAdmin resources page:  
image
  
if the issue hadn't been solved, the entire post of resources would have failed.

PR #897, opened 2024-01-21
tweak(resource/sv_admins): add sv_lan as a reason for Auth failure

Mycroft-Studios:
Currently, having `sv_lan` enabled, will give you a "unknown" reason for txMenu auth failure, this PR simply checks for sv_lan and gives a reason for it :)

### DIFF
--- a/resource/sv_admins.lua
+++ b/resource/sv_admins.lua
@@ -18,7 +18,6 @@ end
 -- Variables & Consts
 local failedAuths = {}
 local attemptCooldown = 15000
-local isLan = GetConvarInt("sv_lan", 0) == 1
 
 -- Handle auth failures
 local function handleAuthFail(src, reason)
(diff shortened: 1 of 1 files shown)


resource/sv_admins.lua
```diff
@@ -18,7 +18,7 @@ end
 -- Variables & Consts
 local failedAuths = {}
 local attemptCooldown = 15000
-
+local isLan = GetConvarInt("sv_lan", 0) == 1
```
wowjeeez: I dont think this should be a statically declared var, since what if someone changes `sv_lan` from the cfx console? Convars can be updated without restarting txadmin.

Mycroft-Studios: in that case, im pretty sure txadmin would break anyway, since it launched in sv_lan and the web side is already populated with sv_lan based information (or, *lack of*, information)

wowjeeez: Its not, there are 0 references to `sv_lan` in a runtime context.

Mycroft-Studios: ah okay, in which case, i will change it to get the value during the event :)

tabarra:
I implemented this a different way (8db72a99c586edc433643a1b3ee9ccb4ae8b690c) because `sv_lan` can change at any time and it being on/off doesn't mean the player that is already connected will or will not have the required identifiers.

PR #900, opened 2024-01-23
Update ar.json

0xAhmedBero:
A Update for ar.json translate fixes


tabarra:
Due to that extra fxmanifest commit, I cherry picked your commit and that would not mark the PR as merged.  
But thanks for the PR!

PR #903, opened 2024-01-25
Update da.json

iplayer1337fivem:
Edited to match the english edition, and rephrased and filled out the blank translations.


tabarra:
Hey @iplayer1337fivem, thanks for the PR!  
I merged it locally because there was an extra `,` breaking the json, and also rebased it.  
Unfortunately this means that the PR will not be marked as merged.

PR #904, opened 2024-01-25
Add missing weapons from DLCs to txAdmin kill internal kill logs

Gravxd:
Adds the missing weapons replacing the current "unknown" that shows in the txAdmin kill logs as well as updates the table containing all the references from GetHashKey to backticks to improve performance.


Mycroft-Studios:
txAdmin currently does not use Lua 5.4, so the backticks will not work.
please change them back to `GetHashKey`

Gravxd:
Small oversight, will change back and re push. Thanks 👍

Gravxd:
Fixed that, bad habit with using Lua 5.4 - all set

tabarra:
Thanks for the PR

PR #906, opened 2024-01-28
locale: fixed cs typo

Marvisak:
A horrible mistake I found while using the txAdmin menu.


nwvh:
damn I was the last one that updated that locale and prolly overlooked this lmao. I wasn't the one that translated this part tho

tabarra:
Thank you for the PR!

PR #909, opened 2024-02-07
Update fxsConfigHelper.ts

mcdnl90:
Fixed conditional that checks if the interface to bind for the FXServer is the same as for txAdmin, which is only mandatory in the case of Zap Hosting because environment requirements. In other scenarios it may make sense that txAdmin binds to a different interface than FXServer does.


tabarra:
This config is not exclusive for zap hosting.

PR #915, opened 2024-02-21
Update AddMasterCallback.tsx

regiments:
correcting the english on the registration page


tabarra:
I thought about using the proper word for it, but I think digits will be more understandable.  
I'm not confident on my decision though.

regiments:
> I thought about using the proper word for it, but I think digits will be more understandable. I'm not confident on my decision though.

Honestly it just confused me a little bit. Characters is a global word, some people will think they can only put numbered combinations which is not only a security risk, its not what the functionality actually is

PR #926, opened 2024-03-16
feat(resource): Return Permissions with Auth

Justus-Braun:
Permissions will get returned with "txAdmin:events:adminAuth" event

### DIFF
--- a/docs/events.md
+++ b/docs/events.md
@@ -156,3 +156,4 @@ Event Data:
 - `netid` (number): The ID of the player or -1 when revoking the permission of all admins (forced reauth).
 - `isAdmin` (boolean): If the player is an admin or not.
 - `username?` (string): The txAdmin username of the admin that was just authenticated.
+- `permissions?` (string[]): The permissions of the admin that was just authenticated.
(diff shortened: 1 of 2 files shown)


tabarra:
Can you describe in details what particular use case you have for this?  
Technically we are planning something different for informing other resources about the admin permissions.

Justus-Braun:
It would allow me to have all permissions and admins be tracked by txAdmin. But a proper Permission System with Groups would be better.

tabarra:
Uhh I actually need an use case for this information.

Justus-Braun:
Sorry, 
I am creating an staff menu for my server which allows staff members to check up with player stats, spawn item, spawn vehicles and more.

When i would get the permissions from the auth event, i would translate txadmin permission to permissions in the staff menu.

This solves the issue when changeing permissions to not do that in 2 places.

But if there is a permission system rework in plan i dont want to interfear with that.

tabarra:
I decided against merging this PR, because if I do so it will create an expectation of stability of the permission names (strings), and I do plan on changing them when rewriting the admin vault and management system.  
Thanks for the PR and suggestion anyways!

PR #927, opened 2024-03-18
Add gold cores to BOOST option for RedM

outsider31000:
This allows to have boost  for players or horses, in RedM they are gold cores.


outsider31000:
sorry my pr is old I didnt check the updated one it seems tit already had this feature.

PR #928, opened 2024-03-18
Getting PlayerPed fails "Somehow" "fix?"

outsider31000:
Somehow GetPlayerPed fails to get a player ped
the other functions break because ped isnt valid.

breaking the loop if ped doesn't exist should Ignore getting  ped info  to avoid more errors

https://cdn.discordapp.com/attachments/704317933353959432/1219098739449528320/image.png?ex=660a112d&is=65f79c2d&hm=6d7b3dc3b31f7b92e39d7b8af8eaf0602503a017247760a61ba2d346f0c32d6f&

Edit: It happens when a player crash, but its random.


outsider31000:
added a new commit to rule out `RegisterKeyMapping` for RedM
image

resource/menu/client/cl_base.lua
```diff
@@ -59,7 +59,7 @@ RegisterCommand('txAdmin:menu:openPlayersPage', function()
   sendMenuMessage('setMenuPage', 1)
   toggleMenuVisibility(true)
   SetNuiFocus(true, true)
-end)
+end,false)
```
BombayV: Any reason for you to set this to false?

resource/menu/client/cl_base.lua
```diff
@@ -194,10 +196,10 @@ RegisterNetEvent('txcl:heal', function()
   local pos = GetEntityCoords(ped)
   local heading = GetEntityHeading(ped)
   if IsEntityDead(ped) then
-    NetworkResurrectLocalPlayer(pos[1], pos[2], pos[3], heading, false, false)
+       NetworkResurrectLocalPlayer(pos[1], pos[2], pos[3], heading, 0, false,false,false)
   end
   ResurrectPed(ped)
-  SetEntityHealth(ped, GetEntityMaxHealth(ped))
+  SetEntityHealth(ped, GetEntityMaxHealth(ped),0)
```
BombayV: SetEntityHealth only requires 2 parameters.

resource/menu/client/cl_base.lua
```diff
@@ -194,10 +196,10 @@ RegisterNetEvent('txcl:heal', function()
   local pos = GetEntityCoords(ped)
   local heading = GetEntityHeading(ped)
   if IsEntityDead(ped) then
-    NetworkResurrectLocalPlayer(pos[1], pos[2], pos[3], heading, false, false)
+       NetworkResurrectLocalPlayer(pos[1], pos[2], pos[3], heading, 0, false,false,false)
```
BombayV: Same thing. NetworkResurrectLocalPlayer does not accept that many parameters.

resource/menu/client/cl_base.lua
outsider31000: yes, as I said in the comment of the commit is for intellisense that requires that parameter to be passed

resource/menu/client/cl_base.lua
outsider31000: it's 3 paremeters , txAdmin isn't just for FiveM
https://rdr3natives.github.io/?_0xAC2767ED8BDFAB15

resource/menu/client/cl_base.lua
outsider31000: txAdmin is not just for FiveM hence why the extra parameters
https://rdr3natives.github.io/?_0xEA23C49EAA83ACFB

tabarra:
This PR got polluted with a bunch of extra changes that I rather not have in the commit.

tabarra:
I noticed that your code is breaking the loop, but we shouldn't do that.  
So I re-implemented as a positive logic instead: 173a3b8c2db47a647baad9cb0d850b0cbb3a4294  
Of course, thanks for the PR and for letting me know of this problem!

Issue #921, opened 2024-03-03
[BUG] Infinite loading screen on Server Log


Keyinator:
**txAdmin/FXServer versions:**
tx: v7.0.0 | fx: b7521
  
**Describe the bug**
You enter an infinite state of the server log loading.

**To Reproduce**
1. Go to 'Server Log'
2. Click on 'View older' (This will put you in mode "LOG")
3. Click on 'View Newer' (This should put you in mode "LIVE" else follow step 3.1)
3.1. (If you're still not on mode "live") Click on 'View Newer'
4. Observe error

**Expected behavior**
Load server log

**Screenshots**
Demo video: https://youtu.be/dkcdR7U_mJ4

**Additional context**
This problem has been observed since pre v7.

tabarra:
Closed by #935, thanks for the report.

PR #935, opened 2024-04-01
[BUG FIX] Server log in web panel will now reconnect when entering live mode

AleksanderEvensen:
Resolves bug report: #921 

The socket was disconnected once checking out previous logs, but was never reconnected upon entering live mode again.
I also moved the event listeners outside the `goLive` function since it would add duplicate listeners on logData.

https://github.com/tabarra/txAdmin/assets/19311856/9affe260-1599-4500-bc87-e378362772ce


tabarra:
Thank you for the PR.

PR #942, opened 2024-04-05
URL Search Params for Players and History page

AleksanderEvensen:
So I checked out your dev notes and found the wouter and URL search params to-do entry, and I thought maybe I could do this. I'm sorry if I was supposed to contact you about this before starting, but giving the feature was pretty simple to implement I figured that I could just make a PR and see if you liked it or not.

The URL parameters will update when the state changes in the component. If this is not the expected behavior and you wanted it to be a copy URL button instead, I could change the code around.

I also removed the export on the `throttleFunc` in both files. Vite HMR won't complain if the export is removed. HMR is also fully functional without the export.

```ts
//FIXME: this doesn't require exporting, but HMR doesn't work without it
// eslint-disable-next-line @typescript-eslint/no-explicit-any, react-refresh/only-export-components
export const throttleFunc = throttle(1250, (func: any) => {
    func();
}, { noLeading: true });
```

Video showcasing the feature.

https://github.com/tabarra/txAdmin/assets/19311856/1c515cd4-e249-4365-8c71-fd49f07eb90c


SirFox1337 (approved):

SirFox1337 (approved):

tabarra:
Hey @AleksanderEvensen, thanks for the PR!  
I decided to implement it myself (2e79d120d2589f6329d1932c12d15988919eade9) because I didn't like some of the changes in your PR, and while investigating how to do it some other way I ended up already re-implementing it.  
Thanks again for the PR and to push me to finally address this.

AleksanderEvensen:
Nice to see the feature finally being implemented. Glad you found another way to solve the problem.

tabarra:
BTW it also supports auto-opening the player and action modals.
```
http://localhost:40120/?playerModal=3VtVF%231
http://localhost:40120/?playerModal=b9503dd10d8d32fc49669c7bb6237a7c79171936
http://localhost:40120/?actionModal=BZET-DK86
```

PR #944, opened 2024-04-09
refactor: death handler

tomson701:
Faster and more optimized method of handling player dead

### DIFF
--- a/resource/cl_logger.lua
+++ b/resource/cl_logger.lua
@@ -339,19 +339,13 @@ AddEventHandler('txAdmin:beta:deathLog', function(ped)
 	processDeath(ped) -- Remember to add a wait function before reviving into an animation.
 end)
 
---[[ Thread ]]--
-local deathFlag = false
-local IsEntityDead = IsEntityDead
-CreateThread(function()
-    while true do
-        Wait(500)
-        local ped = PlayerPedId()
-        local isDead = IsEntityDead(ped)
-        if isDead and not deathFlag then
-            deathFlag = true
-            processDeath(ped)
-        elseif not isDead then
-            deathFlag = false
-        end
-    end
+--[[ Handle death ]]--
+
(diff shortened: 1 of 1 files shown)


Randale-Dienste:
Thats cool

tabarra:
Thanks for the PR  
  
I'm not familiar with those events.  
Would that event also trigger when a death is not from a particular damage? Maybe script-caused death or something.  
Also, why not `CEventDeath` instead?

tomson701:
`CEventDeath` is not being triggered when you death in my case. 
It's handle dead from everything as far as i know. 
Also this event is being used in [ESX Legacy](https://github.com/esx-framework/esx_core/blob/b887903c438c59d04f69d170a7d91892e1386aa2/%5Bcore%5D/es_extended/client/modules/death.lua#L2) and works fine for me.

tabarra:
[Searching on github](https://github.com/search?q=CEventNetworkEntityDamage&type=code) for `CEventNetworkEntityDamage` shows that nobody agrees on what the parameters are.  
Here for me on b2944 the parameter that indicates if the player died is `data[6]`, maybe that's not consistent across game versions?  
  
Also, does this even work on RedM?

AvarianKnight:
Does `CEventPlayerDeath` also not work for this situation? It would be nice to be able to use something that wont be triggered a bunch by other stuff.

Its also probably better to use the specific event as the handler instead of using `gameEventTriggered`

This should work like this:
```lua
AddEventHandler("CEventNetworkEntityDamage", function(data)

end)
```

This got added a [few years ago](https://github.com/citizenfx/fivem/blob/1879d0ca8628ddc80b54ae058f6916a108c3dae2/code/components/citizen-resources-gta/src/GameEventFunctions.cpp#L12C1-L31C5) and be less spammy then `gameEventTriggered`

PR #946, opened 2024-04-21
Update cl_base.lua

Flowrider:
Fixed Error Message RegisterKeyMapping for RedM

RegisterKeyMapping Error TxAdmin

### DIFF
--- a/resource/menu/client/cl_base.lua
+++ b/resource/menu/client/cl_base.lua
@@ -97,7 +97,7 @@ RegisterNetEvent('txcl:setAdmin', function(username, perms, rejectReason)
     debugPrint("^2[AUTH] logged in as '" .. username .. "' with perms: " .. json.encode(perms or "nil"))
     menuIsAccessible = true
     menuPermissions = perms
-    if not IS_REDM then
+    if IS_FIVEM then
         RegisterKeyMapping('txadmin', 'Menu: Open Main Page', 'keyboard', '')
         RegisterKeyMapping('txAdmin:menu:openPlayersPage', 'Menu: Open Players page', 'KEYBOARD', '')
         RegisterKeyMapping('txAdmin:menu:noClipToggle', 'Menu: Toggle NoClip', 'keyboard', '')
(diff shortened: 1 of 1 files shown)


tabarra:
This is a weird issue, never happened before.  
I think this native used to be a no-op in previous client versions.  
Do you know if this bug prevents the tx menu from working in game?

Flowrider:
The menu works perfectly in Game. You only see the error when connecting, which triggers my Monk :D

Actually, it's only an optical bug, at least so far, but it was easy and quick to disable with my change and shouldn't cause any problems.

I don't know from which version this error occurs, I'm currently on Canary, but probably also in beta. I have not tested the release.

outsider31000:
> This is a weird issue, never happened before. I think this native used to be a no-op in previous client versions. Do you know if this bug prevents the tx menu from working in game?

 I made a pr to fix this some weeks ago as well

outsider31000:
#928  specific here with other stuff in the pr too

that native is not available in RedM hence why it errors.

resource/menu/client/cl_base.lua
```diff
@@ -97,11 +97,13 @@ RegisterNetEvent('txcl:setAdmin', function(username, perms, rejectReason)
     debugPrint("^2[AUTH] logged in as '" .. username .. "' with perms: " .. json.encode(perms or "nil"))
     menuIsAccessible = true
     menuPermissions = perms
-    RegisterKeyMapping('txadmin', 'Menu: Open Main Page', 'keyboard', '')
-    RegisterKeyMapping('txAdmin:menu:openPlayersPage', 'Menu: Open Players page', 'KEYBOARD', '')
-    RegisterKeyMapping('txAdmin:menu:noClipToggle', 'Menu: Toggle NoClip', 'keyboard', '')
-    RegisterKeyMapping('txAdmin:menu:togglePlayerIDs', 'Menu: Toggle Player IDs', 'KEYBOARD', '')
-    RegisterKeyMapping('txAdmin:menu:tpToWaypoint', 'Menu: Teleport to Waypoint', 'KEYBOARD', '')
+    if not IS_REDM then
```
tabarra: Better to use `IS_FIVEM` instead.

Flowrider: Done and works.

tabarra:
Just to add context, this actually breaks the menu a little bit.  
User report I got:
> when I give permissions to any one all admin have to relog in to have access to noclip/ tp. baiscily all the features except the TXAdmin tab

Since it fails on the key mapping command, it never reaches the setPermissions.
image

PR #948, opened 2024-05-18
Adds: Mouseover text to player status icons

gegelendvay:
To improve the clarity of what each icon represents, I have added mouseover text (using the title attribute) to each icon. This should make it easier for users to understand the purpose of each icon at a glance.


tabarra:
Thank you for the PR.  
These changes do not work as these elements get translated the `<svg>` tag which does not have the `title` attribute.  
The correct way of implementing this would be to open the igon tag and add a `<title>` element to them, like so:  
```tsx
<ShieldCheckIcon className={cn('h-5',
    rowData.isAdmin ? 'text-warning-inline' : 'text-muted'
)}>
    <title>Is Admin</title>
</ShieldCheckIcon>
```  
  
But with that said, and althought I do see the value in this addition, I will still not apply those changes because the player row must be as light as possible due to how much they are spam-rendered, and the fact that those have been up for over two months and so far nobody asked me to explain what are those, so it's probably _clear enough_ for most users.  
Thanks again for the PR though!

PR #949, opened 2024-05-28
[BUG] RedM fails to authenticate when attempting to assign menu keybinds.

AleksanderEvensen:
When logging in to RedM the below error is displayed and sometimes this causes the menu authentication to fail.
image
image

Executing `/txAdmin-reauth` doesn't fix the problem either, since this will also attempt to assign the menu keybinds.

So i fixed this by adding a check if the menu is running in `FiveM` or `RedM`

### DIFF
--- a/resource/menu/client/cl_base.lua
+++ b/resource/menu/client/cl_base.lua
@@ -97,11 +97,15 @@ RegisterNetEvent('txcl:setAdmin', function(username, perms, rejectReason)
     debugPrint("^2[AUTH] logged in as '" .. username .. "' with perms: " .. json.encode(perms or "nil"))
     menuIsAccessible = true
     menuPermissions = perms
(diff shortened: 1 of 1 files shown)


outsider31000:
this has already been added. you need to wait for new update
https://github.com/tabarra/txAdmin/commit/4d388f82f32a4f6ec2e889272d5ecd55e93a2200

tabarra:
Thanks for the contribution @AleksanderEvensen, but as outsider said, this one was actually fixed in #946.  
I tend to only merge requests when I'm ready to release new updates to prevent this from happening, but this time some plans changed last minute so here we are :P  
  
But thanks again!

AleksanderEvensen:
Aaah. My bad I didn't check the other PRs to see if it was already fixed.

PR #953, opened 2024-06-10
(feat/new-perf): Fixed bug where it would return incorrect undefined

AntonsWorkshop:
Fix "getMinTickIntervalMarker" returning undefined when it's not supposed to.


AntonsWorkshop:
I was able to fix the issue where `getMinTickIntervalMarker` would return a wrong undefined. 

This still leaves `expect(result).toBeUndefined();` returning undefined.

PR #956, opened 2024-06-11
(fix) Removes lipsu from player crashes page.

ignBiometrical:
Self explanatory, just removes lipsu from the crash page.


Willow-6855:
it is a easter egg not a bug.

tabarra:
I got it, thank you very much.

PR #961, opened 2024-06-13
Update Arabic locale

mash6oob:
I changed some labels that looked like it was google translated, I only changed the ones that are obvious.


tabarra:
Thanks for the PR!  

PR #965, opened 2024-06-29
Fix DA Locales

DetErDugget:
Some translations in da.json was wierd.


tabarra:
Thanks for the PR!  

PR #968, opened 2024-07-22
add unique ID to "empty name"

itsRVX:
This PR adds a simple 4 digit unique ID to each "empty name" 

 E.g... empty name F8B2

### DIFF
--- a/shared/cleanPlayerName.ts
+++ b/shared/cleanPlayerName.ts
@@ -1,4 +1,4 @@
-import { randomBytes } from 'node:crypto';
+import crypto from 'node:crypto';
 
 const byteToRegexHex = (byte: number) => {
     return '\\u' + byte.toString(16).padStart(2, '0').toUpperCase();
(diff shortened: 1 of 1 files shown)


tabarra:
I see what you are trying to do, but here's the issue: there is the possibility that a legitimate name ends up being considered "empty name" due to how I'm processing the string, and then if you wanna search for that same name, you need to find it without any randomness.  
  
So maybe what should be done instead is adding a hash-based unique identifier instead of pure random?  
But still it would not match partial names.... I'm not sure what is the best approach. Do you have any ideas?

tabarra:
Maybe instead of random, what if the names get encoded to `emptyname${dropped}`, where `dropped` is the hex representation of first 6 characters? In this case searching will still try to map the start of the search string.  
Maybe even drop the "emptyname" and just replace by some unicode symbol for "invalid stuff ahead".

tabarra:
Thanks for the commit.  
What if instead of empty, we just add `∅${dropped_as_hex}`?  
The symbol U+2205 is "Empty Set", which makes sense.  
Let me know what you think.  
I'm less concerned of code writing and more concerned with the 💡🧠

itsRVX:
Ah, so then the sanitized displayName in the event of no length would be something like "∅C0A6"

tabarra:
Yes, what do you think?

itsRVX:
Works for me. As long as they are unique I'm good for it. I'll make the changes

tabarra:
The issue with hashing is that you cannot search for a partial name, just the full name.  
So I did something different, I rewrote the entire `cleanPlayerName()` method to handle things differently, and when the name ends up being empty after any of the steps, it gets converted to its code points, which can be searched by typing in the code point or searching for the character, in which case tx would convert it and search by the codepoint represented in hex.  
I believe this is the best approach for this problem.  
Commit: 55d2282366f34ff5dfca3e5910665fbed20d3172  
  
In the example below, I ran `window.invokeNative('changeName', '\u{1160}\u{3164}');` in the mpMenu devtools to set the name to two invisible characters:  [HANGUL JUNGSEONG FILLER](https://unicode-explorer.com/c/1160) and [ㅤHANGUL FILLER](https://unicode-explorer.com/c/3164) (previously not cleaned up by the method), and here is the result:  
  
server log
search by codepoint
search by char

tabarra:
Thanks for the PR though!

tabarra:
Quick Update: now this also works for the in-game playerlist.  
  
in-game menu playerlist
chat command

PR #969, opened 2024-07-26
Update de.json

PC1up:
Translate the filters and one line that wasn't translated as well


tabarra:
As always, thanks for the PR!  

PR #975, opened 2024-08-15
feat(panel/core): add a gamebuild option

spacevx:
Someone told me it was pointless to do it because the legacy pages will be rewritten, but I already did it, so I'm sharing it.
It's just an option to change the gamebuild of the FiveM client. I didn't do it for RedM (for now).
If something is wrong, don't hesitate to let me know.

### DIFF
--- a/core/components/ConfigVault.js
+++ b/core/components/ConfigVault.js
@@ -162,6 +162,7 @@ export default class ConfigVault {
                 commandLine: toDefault(cfg.fxRunner.commandLine, null),
                 logPath: toDefault(cfg.fxRunner.logPath, null), //not in template
                 onesync: toDefault(cfg.fxRunner.onesync, 'legacy'),
+                gamebuild: toDefault(cfg.fxRunner.gamebuild, null),
                 autostart: toDefault(cfg.fxRunner.autostart, null),
                 restartDelay: toDefault(cfg.fxRunner.restartDelay, null), //not in template
                 shutdownNoticeDelay: toDefault(cfg.fxRunner.shutdownNoticeDelay, null), //not in template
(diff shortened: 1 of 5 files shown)


tabarra:
Thanks for the pull request, that was very thorough of you for even including the cfg validator and launch parameter changes.  
Unfortunately I cannot merge this PR, at least not for now.  
The issue is that txAdmin and fxserver versions are not strictly bound together, which means the list is bound to be wrong eventually, even if we ignore the fact that some people backport tx to older artifacts.  
  
Some time ago I was talking to Cfx about it, and there are plans for them to implement a new native to get the list of all game builds available for the current fxserver version (ref: CFX-738), but there is no ETA for that to be implemented. So for now, this cannot be merged.  
  
I will leave this PR open though, if cfx releases the native before I migrate the settings page, I'll add to it and merge, otherwise will end up closing it.  
Thanks again for the contribution, though.

spacevx:
Hello, I can implement a workaround for this. I just need to find the artifact version based on each game build (Windows/Linux) and compare with the actual artifact. Even if CFX releases a native, it means older server artifacts won’t be compatible since the native will only be available on newer artifacts

tabarra:
Additional considerations from discord convo:  
  
> Not having the feature right now is not stopping anyone from using any game build available, and yet, if I merge the pr, two issues will happen:
> 1. If a new DLC becomes available and i'm working on some update, people will want me to release a new txAdmin update just so they can use the new DLC, and even if i'm willing to release said update, then tens of thousand of admins will get a notification to update, but for an update that doesn't add anything meaningful for them.
> 2. Or worse, people will see the changelog and the new dlc available, and will try it even if it's not really ready for prime time just yet.

tabarra:
Thanks again for the PR, but the old settings page is now gone.  
Also, although I'm still considering adding that option to txAdmin (probably not the settings page), this requires some further thinking with the cfx team.

PR #976, opened 2024-08-16
fix(streaming): release ptfx asset after use

Geden420:
The PTFX asset used in the createPlayerModePtfxLoop() function is never released after use. Note that the 'core' PTFX asset alone streams 21.95 MiB.


tabarra:
Hmm, is that a memory leak?  
Does every use of `RequestNamedPtfxAsset(PTFX_DICT)` increase the memory by 22mb?

Geden420:
It only increases streaming the first time it requests. Memory is not changed.

But if it is not released the script will keep streaming it

Before changes. After streaming (idel)
image

After changes. After streaming (idel)
image

tabarra:
Ahh I see... would remove the effect cause apparent lag on each use?

Geden420:
I haven't personally experienced any problems, but I do know that unreleased assets can cause streaming issues.

Geden420:
If the asset request gets spammed, it could potentially cause some lag while it's being used? But the usecase is the same with troll actions when the animal model asset is requested and then released using SetModelAsNoLongerNeeded().

However, I'm not familiar enough with FiveM streaming to provide a definitive answer.

tabarra:
Maybe we should start a timer to reset it after 120s, and each use resets the timer.

Geden420:
I don't think that's necessary. I just tested spamming with and without release, and I didn't notice any change in lag spikes or FPS drops. However, keep in mind that it's a client-sided resmon.

With changes:
https://streamable.com/q1vf0o

Without changes:
https://streamable.com/06b3qg

tabarra:
Ohh streamable... `This video isn't available anymore`.

Mycroft-Studios:
> Maybe we should start a timer to reset it after 120s, and each use resets the timer.

Personally, I think this would be the better way, since it uses the effect with toggling both on and off, and since people mainly use noclip to get to somewhere fast, it makes sense to hold onto it, when its going to be used within the next minute or so anyway, and then release after that.

Geden420:
Or use the network version of `StartParticleFxLoopedOnEntity` and mix it with some of that magic. Then we can make the code only request assets and run for the user switching mode?

https://docs.fivem.net/natives/?_0x6F60E89A7B64EE1D

tabarra:
Thanks for the PR.

PR #981, opened 2024-08-30
Update translation for Vietnameses

nkyo:
Update translation for Vietnameses


tabarra:
Thank you very much for the PR.  
Later today I will do a commit adding a new key to `nui_warning`, feel free to make a new PR to translate that as well.

PR #982, opened 2024-09-02
Update de.json

furcoite-nirual:
translated the `kick_messages`


tabarra:
Thank you for the PR!

PR #986, opened 2024-09-23
Added Nepali Locale

RojinxChhetri:
This pull request adds complete Nepali language support to our server management UI. The entire JSON localization file has been translated from English to Nepali, enabling native Nepali speakers to use the interface in their preferred language.


tabarra:
Thanks for the PR, but you will need to change a few things before I can merge it.  
First, make sure to read the docs on it [translating.md](https://github.com/tabarra/txAdmin/blob/master/docs/translation.md), then rebase your file on this updated [en.json](https://github.com/tabarra/txAdmin/blob/develop/locale/en.json).  
Then make sure to use `ne` instead of `np`, as that appears that ne is the Nepali [ISO 639-1](https://en.wikipedia.org/wiki/ISO_639-1) code.  
  
For comparison, these are the errors raised by the `npm run locale:check` command:
- $meta.humanizer_language: language not supported
- kick_messages.everyone: missing key
- kick_messages.player: missing key
- kick_messages.unknown_reason: missing key
- nui_warning.stale_message: missing key
- nui_warning.dismiss_key: missing key

tabarra:
Apparently humanize-duration doesn't support Nepali, so you will need to leave that as `en`

RojinxChhetri:
> Apparently humanize-duration doesn't support Nepali, so you will need to leave that as `en`

So We can't use Nepali Locale at the moment?

RojinxChhetri:
> You can, but you need to set `$meta.humanizer_language` to `en`.

Oh ok got it working on it thanks for it.

RojinxChhetri (commented):
Please have a look and let me know if anything else need to be changed.

tabarra:
Still these pending changes:    

[pt.json] Errors found in Nepali locale:
- kick_messages.everyone: must contain the placeholders ["%{reason}"]
- kick_messages.player: must contain the placeholders ["%{reason}"]
- nui_warning.instruction: must contain the placeholders ["%{key}","%{smart_count}","||||","%{key}","%{smart_count}"] 

And if possible move the lines from in the locale map up to match the alphabetical sorting order.

tabarra:
Ahh! I just noticed the error message was a bit misleading, you need to remove the `${name}` from the translation `xxxxx`.  
Maybe something like this would work?  
```
"kick_messages": {
    "everyone": "सबैलाई निकालियो कारण: %{reason}।",
    "player": "तिमीलाई निकालियो कारण: %{reason}।",
    "unknown_reason": "अज्ञात कारण"
},
```

RojinxChhetri:
> Ahh! I just noticed the error message was a bit misleading, you need to remove the `${name}` from the translation `xxxxx`. Maybe something like this would work?
> 
> ```
> "kick_messages": {
>     "everyone": "सबैलाई निकालियो कारण: %{reason}।",
>     "player": "तिमीलाई निकालियो कारण: %{reason}।",
>     "unknown_reason": "अज्ञात कारण"
> },
> ```

Yeah this also makes sense and also better. Let me quickly update it.

RojinxChhetri (commented):
Updated the purposed changes

PR #994, opened 2024-10-15
Update pt.json

Legen141215:
Updated pt.json from brazilian to portuguese and changed some fields from english to portuguese aswell.


tabarra:
I'm sorry, but long long time ago it was decided there would only be one pt translation file (after we tried having both pt-BR and pt-PT), and since Brazilians outnumber 21 to 1, we decided to keep it in Brazilian Portuguese.  
Feel free to use this as a custom locale file as described in the docs.

PR #996, opened 2024-10-16
fix(server): many changes

Zignixx:


tabarra:
Hi @Zignixx 

I noticed you've forked **txAdmin** and renamed it to `ImmortalRoleplay` in many (all?) public facing messages, which, honestly, feels like a low blow in the open-source world.  
  
What is more pressing, though, is the removal of the public-facing copyright notice from the web UI.  
Per the MIT license, the original copyright notice has to remain in all copies or substantial portions of the software. Removing it isn't compliant with the license terms. **Please restore the copyright notice and the link to the _original_ LICENSE file in your fork.**  
Let me know if you need any clarification on this.

_(ps. Since issues are disabled in your fork, I was compelled to comment in this PR instead)_

PR #1007, opened 2024-10-29
chore(locales): Update nl translations

mardev-xyz:


tabarra:
Thanks for the PR!

yorick2002:
Correct grammar would be "gekickt" instead of "gekicked" @mardev-xyz

PR #1008, opened 2024-11-05
fix(panel): don't focus bg color on select dropdownmenu in desktopnavbar

Misha0717:
If you click on the dropdownmenu and move your mouse away of the dropdownmenu the focus is still activated on the button, this means you can have 2 buttons with the focus attribute in the desktopnavbar.

Example:
image

I couldn't find a way to disable the focus attribute for the specific button, but for now the 'NavigationMenuTrigger' is only used in the desktopnavbar, thats why I removed the full 'focus:bg-secondary'.


tabarra:
Thank you.

PR #1018, opened 2024-12-31
feat(locales): add/finish Latvian translation

NotReloads:
Updated the latvian translation to suit more + missing translations.


tabarra:
Thanks for the PR!

PR #1020, opened 2025-01-02
feat(deployer): add task retry mechanism with improved error feedback

RicardoFaisca12:
# Task Retry System for Deployer

As someone who has set up multiple FiveM servers, I've often run into issues where the deployment would fail due to temporary problems like:
- Slow internet causing download timeouts
- GitHub API rate limits
- Database connection hiccups
- Random network issues
- File system locks

This was super frustrating because every time it happened, I had to start the entire deployment process from scratch. 

So I added a retry system that:
1. Automatically retries failed tasks up to 3 times
2. Uses smart delays between retries (1s, 2s, 3s)
3. Shows clear feedback about which retry attempt it's on
4. Only fails the deployment if all 3 attempts fail

## Example Output
```log
[12:34:56] Running [task1:download_file]... ❌
[12:34:56] Task failed (retrying 1 of 3): Connection timed out
[12:34:57] Running [task1:download_file]... ❌
[12:34:57] Task failed (retrying 2 of 3): Connection timed out
[12:34:59] Running [task1:download_file]... ✔️

### DIFF
--- a/core/extras/deployer.js
+++ b/core/extras/deployer.js
@@ -1,30 +1,33 @@
-const modulename = 'Deployer';
-import path from 'node:path';
-import { cloneDeep } from 'lodash-es';
-import dateFormat from 'dateformat';
-import fse from 'fs-extra';
-import open from 'open';
-import YAML from 'js-yaml';
-import getOsDistro from '@core/extras/getOsDistro.js';
-import { txEnv } from '@core/globalData';
-import recipeEngine from './recipeEngine.js';
-import consoleFactory from '@extras/console';
+const modulename = "Deployer";
+import path from "node:path";
+import { cloneDeep } from "lodash-es";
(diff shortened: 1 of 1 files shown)


tabarra:
Hey Ricardo, thanks for the PR but I cannot accept it due to:
1. It targets master, which is severely outdated compared to the develop branch.
2. It reformats the entire file, which makes it hard for me to track what are the intended changes.
3. We cannot retry tasks, without being able to rollback a failed task which might or might not have messed with the files or the deployer context state.
4. If you are specially worried about the download task, you should keep in mind it already retries the download 5 times. Increasing that to 15 is unlikely to make a difference.
5. The recipe system is likely going to get refactored in the upcoming months to download a pre-built recipe instead of building it all from scratch, so the issues with tasks failing should be resolved.

Thanks again for the PR.

PR #1023, opened 2025-01-09
fix: Change keymaster to portal.

SnepCnep:
The 'https://keymaster.fivem.net/' will soon be ended and switch to the new panel "portal.cfx.re"


SnepCnep:
Close it of the dupe. ;)

PR #1025, opened 2025-01-17
Server event to update mismatched vehicle types list

Scullyy:
- Adds server event to update the mismatched vehicle types list, this removes the requirement to update the list every time you update your artifacts.
- Moves mismatched types to server due to being required for the event to work, event needs to be server sided to prevent abuse.
- Tested on b3258 with the latest artifacts and txAdmin version as of this PR and working.

Example usage:

```lua
TriggerEvent('txAdmin:menu:addMismatchedTypes', {
    caddy3 = 'automobile',
    trailersmall2 = 'trailer',
    scrap = 'automobile',
    tractor3 = 'automobile'
})
```

### DIFF
--- a/resource/menu/client/cl_vehicle.lua
+++ b/resource/menu/client/cl_vehicle.lua
@@ -22,74 +22,16 @@ local function getPedVehicle()
     end
 end
 
--- NOTE: this is not a complete list, but most others have the type "automobile"
-local vehClassNamesEnum = {
-    [8] = "bike",
-    [11] = "trailer",
-    [13] = "bike",
-    [14] = "boat",
-    [15] = "heli",
-    [16] = "plane",
-    [21] = "train",
-}
-
--- Since we don't have the vehicle types on the server, we need this translation table
--- NOTE: this list was generated for game build 2802/mpchristmas3
--- How to update the list: https://gist.github.com/tabarra/32ef90524188093ab4218ee7b5121269
-local mismatchedTypes = {
-    ["airtug"] = "automobile",       -- trailer
-    ["avisa"] = "submarine",         -- boat
-    ["blimp"] = "heli",              -- plane
-    ["blimp2"] = "heli",             -- plane
-    ["blimp3"] = "heli",             -- plane
-    ["caddy"] = "automobile",        -- trailer
-    ["caddy2"] = "automobile",       -- trailer
-    ["caddy3"] = "automobile",       -- trailer
-    ["chimera"] = "automobile",      -- bike
-    ["docktug"] = "automobile",      -- trailer
-    ["forklift"] = "automobile",     -- trailer
-    ["kosatka"] = "submarine",       -- boat
-    ["mower"] = "automobile",        -- trailer
-    ["policeb"] = "bike",            -- automobile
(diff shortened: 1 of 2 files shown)


tabarra:
@Scullyy Thanks for the PR.

If the system is broken in txAdmin, the right thing is for it to be fixed in txAdmin instead of getting individual server owners to fix it themselves.  
  
That being, said, I tried updating this using [my old guide](https://gist.github.com/tabarra/32ef90524188093ab4218ee7b5121269) and something is very wrong.  
The list of classes seem to now be 22 items long instead of just 7, as can be seen in the [fivem docs](https://docs.fivem.net/natives/?_0x29439776AAA00A62), and using this list marks all 844 vehicles as divergent.  
But furthermore, I tried spawning a vehicle from the list and the game didn't crash as it usually did.  
  
Can someone that knows more about the game help me fix this issue? Is this still even an issue?!

Scullyy:
> @Scullyy Thanks for the PR.
> 
> If the system is broken in txAdmin, the right thing is for it to be fixed in txAdmin instead of getting individual server owners to fix it themselves.
> 
> That being, said, I tried updating this using [my old guide](https://gist.github.com/tabarra/32ef90524188093ab4218ee7b5121269) and something is very wrong. The list of classes seem to now be 22 items long instead of just 7, as can be seen in the [fivem docs](https://docs.fivem.net/natives/?_0x29439776AAA00A62), and using this list marks all 844 vehicles as divergent. But furthermore, I tried spawning a vehicle from the list and the game didn't crash as it usually did.
> 
> Can someone that knows more about the game help me fix this issue? Is this still even an issue?!

Hey, nothing was broken in TxAdmin. The vanilla vehicles were just used an example to show usage, this is just so server owners can add their addon vehicles without needing to update the list each time but the current implementation works perfectly fine for vanilla vehicles.

tabarra:
It's weird though, because the data changed, and it still works without updating it?!

Scullyy:
> It's weird though, because the data changed, and it still works without updating it?!

Some of the newer dlc vehicles may not work as they’re not listed but I meant the old method can continue to be used for default GTA vehicles with my PR as all I did was just move it to the server and add an api to add to the list externally so that people can put their addon vehicles that they download for their server on the list and they won’t need to make any changes between TxAdmin updates. 

For example I have a hot air balloon vehicle and a truck that has a mismatched type so I would trigger the event in a server script which will put them on the mismatched list and allow them to spawn correctly in TxAdmin between updates without needing to edit the monitor resource.

tabarra:
When trying to spawn a vehicle with wrong class/type, does it crash the client like it used to?

Scullyy:
> When trying to spawn a vehicle with wrong class/type, does it crash the client like it used to?

When I spawn a vehicle that has a mismatched type (without it being on the list to correct them) it will spawn an incorrect vehicle. Like let's say a truck was a trailer instead of spawning the truck it would spawn a trailer.

Scullyy:
But to be clear this PR wasn't to resolve any kind of bug as I had no issue with the current system, this is only to implement an API so you can put your add-on vehicles onto the list that you download for your server without needing to edit the monitor resource each time.

PR #1027, opened 2025-01-23
tweak(core): use si for process data

million1156:
## This PR is still a draft. I am not sure if exec is the best possible implementation, but it seems okay enough to create a draft PR.

This PR fixes an issue that was introduced with Windows 11 & Windows Server 2025, which removed `wmic`. 

Instead of using the deprecated command, we instead call `Get-WmiObject`, whose documentation can be found [here](https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.management/get-wmiobject?view=powershell-5.1). The command we use to get the data is: 
```ps1
Get-WmiObject -Query "SELECT * FROM Win32_Process WHERE ProcessId = ${fxsPid}" |  Select-Object -ExpandProperty WorkingSetSize
```
, where:

- fxsPid is the fxServer process
- WorkingSetSize is the amount of memory fxServer is currently occupying.

Essentially, we're given the amount of memory (in bytes) that the process is currently consuming. So, we do the same division that we used to use for Linux processes.

The only concern I have here is that it *may* not be the best idea to use `exec`. However, I've tried my best to prevent any security issues. I'm curious if there's any better ways to go about this! But it seems like the most viable solution as of right now.


tabarra:
Hey @million1156, thanks for the (draft) PR.

txAdmin uses a bunch of different ways to get a bunch of different host data, like windows / distro name, processor speeds, processes memory usage, etc. And all of that was built individually without much consideration for the overall.  
  
More recently I've thinking of just using thinking of refactoring those parts of code to use the `systeminformation` library.  
In the past I had avoided it because in some systems, even getting the processes memory usage _could_ cost about 10 seconds of the thread going 100% usage, but now I think it's probably the time for the change.  
  
Here is a copypaste from the devnotes about this subject:

- [ ] Try to replace all the host stats/data with stuff from the SI lib (eg `systeminformation.processLoad()`).
    - They are already using GWMI: https://github.com/sebhildebrandt/systeminformation/issues/616
    - Pay attention to the boot and shutdown comments
    - NOTE: keep in mind the processor time vs utility difference:
        - https://github.com/citizenfx/fivem/commit/034acc7ed47ec12ca4cfb64a83570cad7dde8f0c
        - https://learn.microsoft.com/en-us/troubleshoot/windows-client/performance/cpu-usage-exceeds-100

million1156:
Hey @tabarra,

Thanks for the quick response! I've created a new commit that now uses `si`. It's a little more convoluted to get working, but it works in my testing :) Let me know what you think!

EDIT: In the future I want to try and replace *every* stats, but I think this is a great start since this currently affects anyone who's using Windows Server 2025 or Windows 11 (i.e. a dev hosting a server from their home computer/server)

tabarra:
Note to future self:  
The call to `execa()` inside `startReadyWatcher.ts -> getOSMessage() -> getOsDistro() -> windowsRelease()` on Bun makes the `process.on()` signals (SIGINT, SIGTERM, etc) event handlers registered on `setupProcessHandlers.ts` to not fire anymore.  
Replacing that call with `si.osInfo()` seems like a viable option.

PR #1030, opened 2025-02-10
Modify translate Vietnameses

nkyo:
Modify translate Vietnameses


tabarra:
Thank you for the PR!

PR #1051, opened 2025-05-03
Option to hide warning when button is disabled.

Qotrek:
In my case, the player who received the warning and was handcuffed couldn't hide it because the handcuff script blocked the control from jumping. :D

### DIFF
--- a/resource/cl_main.lua
+++ b/resource/cl_main.lua
@@ -71,7 +71,7 @@ RegisterNetEvent('txcl:showWarning', function(author, reason, actionId, isWarnin
         local count = 0
         while true do
             Wait(100)
-            if IsControlPressed(dismissKeyGroup, dismissKey) then
+            if IsControlPressed(dismissKeyGroup, dismissKey) or IsDisabledControlPressed(dismissKeyGroup, dismissKey) then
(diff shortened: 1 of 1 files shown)


yorick2002:
~~The point of the warning is that it's annoying to the offender 🤷‍♂️~~ nvm i misread my bad

tabarra:
Ohh so you are just double checking if the key is pressed, even if it's disabled.  
Got it, thanks for the PR, will see if someone can test so I can merge for the next update.

resource/cl_main.lua
```diff
@@ -71,7 +71,7 @@ RegisterNetEvent('txcl:showWarning', function(author, reason, actionId, isWarnin
         local count = 0
         while true do
             Wait(100)
-            if IsControlPressed(dismissKeyGroup, dismissKey) then
+            if IsControlPressed(dismissKeyGroup, dismissKey) or IsDisabledControlPressed(dismissKeyGroup, dismissKey) then
```
csskroubledev, line 74: No need to call this twice, will still work

```suggestion
            if IsDisabledControlPressed(dismissKeyGroup, dismissKey) then
```

PR #1056, opened 2025-06-02
locale: re-translate norwegian

MadsLeander:
The norwegian localization lacked a substantial amount of translations, and the translations that were present were of a sub-optimal quality (weird sentence structure, wrong usage of "et" and "en", etc.). I therefore decided to re-translate the entire menu.

Here are some decisions I've noted down while translating:
The `$meta.humanizer_language` was wrong? The postfix `(Bokml)` should be `(Bokmaal)` as it is written `Bokmål` in Norwegian, "Å" is always written as double a (AA) when Å is not available/allowed. There is also an argument to be made that the post-fix should be removed, as there isn't any version for Nynorsk present, and I doubt there ever will be. However, to not acidently screw up some standard I've desided to leave it be.

I've "translated" restart as "restart" instead of a more classical "omstart". Both are valid translations, and the only reason that I landed on "restart" is that it is more commonly used among the tech-savvy and gaming communities.

I've also not found any good translation for "Spawn" in norwegian (gamers just use the word "spawn"), so I've just left it as "Spawn" for now. I also see that the danish localization does the same.

As requested in the translation.md, here are some in-game images of the menu using the new translations.
Skjermbilde 2025-06-02 134340
Skjermbilde 2025-06-02 134359
Skjermbilde 2025-06-02 134415
Skjermbilde 2025-06-02 134430
Skjermbilde 2025-06-02 134453

### DIFF
--- a/locale/no.json
+++ b/locale/no.json
@@ -1,366 +1,366 @@
 {
     "$meta": {
-        "label": "Norwegian (Bokml)",
+        "label": "Norwegian (Bokmaal)",
         "humanizer_language": "no"
     },
     "restarter": {
-        "server_unhealthy_kick_reason": "serveren må startes på nytt, vennligst koble til på nytt",
-        "partial_hang_warn": "Due to a partial hang, this server will restart in 1 minute. Please disconnect now.",
-        "partial_hang_warn_discord": "Due to a partial hang, **%{servername}** will restart in 1 minute.",
-        "schedule_reason": "et omstart er planlagt kl. %{time}",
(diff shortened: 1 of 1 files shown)


MadsLeander:
I just saw that I targeted the wrong branch. Thanks for updating it!

tabarra:
Thanks for the PR!  
Now there is another error:
```
Checking validity of the locale files based on 'en.json'.
[no.json] Errors found in Norwegian (Bokmaal) locale:
- ban_messages.reject.label_expiration: untrimmed string
```

MadsLeander:
Not sure how I managed to miss that. But it should be fixed now.

PR #1058, opened 2025-06-15
feat(server): add RedM explosion types

outsider31000:
txAdmin is missing RedM explosion types leading to confusion, this change allows to use proper explosion names for RedM


tabarra:
Hey @outsider31000, are these values still correct with latest GTA and RDR game builds?

tabarra:
**Note to future self:**
- when merging, do a var for fivem, one for redm, and then ternary to pick the right one, as is being done in the rest of the codebase.
- due to the unwanted changes, might be worth to just commit and add outsider as co-author

outsider31000:
for RDR I can say it is.

PR #1059, opened 2025-06-16
fix(menu) Prevent players being able to teleport to themselves

yorick2002:
This PR adds a check to the "GO TO" option in the menu to prevent users from teleporting to themselves 

Preview:

https://r2.fivemanage.com/KFTFxoppLJfPvAgnQBsLL/teleportToSelf.gif

### DIFF
--- a/resource/menu/client/cl_main_page.lua
+++ b/resource/menu/client/cl_main_page.lua
@@ -58,6 +58,13 @@ end)
 
 
 RegisterSecureNuiCallback('tpToPlayer', function(data, cb)
+    local targetServerId = tonumber(data.id)
+
+    -- check for self-teleport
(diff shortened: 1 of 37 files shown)


resource/menu/client/cl_main_page.lua
```diff
@@ -58,6 +58,13 @@ end)
 
 
 RegisterSecureNuiCallback('tpToPlayer', function(data, cb)
+    local targetServerId = tonumber(data.id)
+
+    -- check for self-teleport
+    if targetServerId == GetPlayerServerId(PlayerId()) then
+        return sendSnackbarMessage('error', 'nui_menu.player_modal.actions.interaction.notifications.teleport_yourself', true)
```
mikigoalie, line 65: doesn't chromium / nui always expect a cb to be invoked?

PR #1060, opened 2025-06-28
fix(menu) Fix playernames with invalid chars in name

Misha0717:
there are some chars when you have that in your name, the name will no longer be shown at all.
The pr will remove the invalid char.

**Preview:**
image

**Player name:**
image

### DIFF
--- a/resource/menu/client/cl_player_ids.lua
+++ b/resource/menu/client/cl_player_ids.lua
@@ -130,7 +130,8 @@ local function showGamerTags()
             or not IsMpGamerTagActive(playerGamerTags[pid].gamerTag)
         then
             local playerName = string.sub(GetPlayerName(pid) or "unknown", 1, 75)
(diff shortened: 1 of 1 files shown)


tabarra:
What's the definition of `%w`? Sounds a bit vague and not friendly towards scripts that are not alphabetic (Thai, Farsi, Cyrillic, Chinese, etc.).
Ideally we should use the txAdmin `cleanPlayerNames()` to get the name to be displayed, but the best implementation would be for Cfx to implement a `UpdatePlayerName(tempNetId: number, name: string)` that could be called by txAdmin at the handling of `playerConnecting()`, as well as allow other resources to set the player's rank, RP name, etc.

Misha0717:
I have tested it with chinese chars, and as you told unfortunately it will remove also that kinda chars. The only thing to fix this is prob gsub specific chars like in the preview seen. Would be beter ofc if cfx fix this.

PR #1071, opened 2025-09-11
feat(discord): implement Discord API helper for member and role resolution

jxnnyo:
## Summary
Implements a lightweight Discord API helper that uses direct HTTP requests instead of requiring a full Discord bot client, allowing Discord whitelist functionality without token conflicts.

## Problem Solved
Previously, using Discord whitelist features required `discordBot.enabled = true`, which created a Discord.js client with websocket connection. This caused token conflicts when users already had their own Discord bot running on the same token, since Discord only allows one bot connection per token.

## Solution
- **New Discord API Helper** (`core/modules/DiscordBot/discordApiHelper.ts`)
  - Uses raw Discord REST API calls via HTTP instead of websocket connection
  - Handles member resolution by ID with role checking
  - Includes 1-minute caching to reduce API calls
  - Only activated when Discord whitelist modes are enabled
  - No persistent connection - only makes requests when needed

## Changes
- **Enhanced Player Join Checks** (`core/routes/player/checkJoin.ts`)
  - Integrated Discord API helper for whitelist validation
  - Uses HTTP API calls instead of Discord.js client methods
- **Settings Configuration** (`core/routes/settings/saveConfigs.ts`)
  - Enhanced Discord configuration validation
  - Better handling of Discord API helper initialization
- **Frontend Whitelist UI** (`panel/src/pages/Settings/tabCards/whitelist.tsx`)
  - Minor UI improvements for Discord whitelist configuration

## Key Benefits
- **No Token Conflicts**: Users can run their own Discord bots while using txAdmin's Discord whitelist
- **Lightweight**: Only makes HTTP requests when checking player joins
- **Backwards Compatible**: Existing Discord whitelist configurations continue working
- **Better Performance**: Cached member lookups reduce redundant API calls

## Technical Details
The helper automatically activates when `whitelist.mode` is set to `discordMember` or `discordRoles` and Discord credentials are configured, regardless of `discordBot.enabled` status. This allows Discord whitelist functionality without requiring the full bot client.

### DIFF
--- a/core/modules/DiscordBot/discordApiHelper.ts
+++ b/core/modules/DiscordBot/discordApiHelper.ts
@@ -1,0 +1,175 @@
+const modulename = 'DiscordApiHelper';
+import consoleFactory from '@lib/console';
+const console = consoleFactory(modulename);
+
+const DISCORD_API_BASE = 'https://discord.com/api/v10';
+
+export type DiscordMemberResult = {
+    isMember: boolean;
+    memberRoles?: string[];
+    error?: string;
+};
+
+export type DiscordUserProfile = {
+    tag: string;
+    avatar?: string;
+};
+
+/**
+ * Lightweight Discord API helper for member/role resolution
+ * Uses HTTP API directly without requiring a full Discord bot client
+ */
+export class DiscordApiHelper {
+    private token: string;
+    private guildId: string;
+    private headers: Record<string, string>;
+    private guildName?: string;
+    private memberCache = new Map<string, { timestamp: number; data: DiscordMemberResult }>();
+    private readonly CACHE_TTL = 60_000; // 1 minute cache
+
+    constructor(token: string, guildId: string) {
+        this.token = token;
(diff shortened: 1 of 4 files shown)


tabarra:
> This caused token conflicts when users already had their own Discord bot running on the same token, since Discord only allows one bot connection per token.

I'm sorry, but this is a problematic solution for a made-up problem.  
This is like saying "I can't use my glass to drink milk because I'm already using it to drink Pepsi from"... well then, get another cup?!  
**In other words: just make a new bot.**  
  
Further reasons to close this PR:  
- HTTP requests at every player join is a great way to get your server IP rate limited by Discord, especially when on shared hosting.
- Bad separation of concerns, all the role resolution logic should be contained by the DiscordBot module, and you have leaked it into `core/routes/player/checkJoin.ts`.
- **`fetch` is not globally defined, meaning your code literally does not work.** Did you use AI to write this PR?!

PR #1074, opened 2025-09-19
feat(discord) Add maintenance mode status to discord bot embed

yorick2002:
This PR aims to add a "Maintenance" status to the discord bot embed when the server is in admin-only mode.


MacLove13 (approved):

tabarra:
Need to figure out how to "inject" the new JSON params into the configs, even if the props are technically optional

PR #1076, opened 2025-09-20
docs(events): fix wrong description for actionId in actionRevoked event

cxllin2024:
This PR fixes the description of "**actionId**" in the "**actionRevoked**" event and also removes trailing whitespace.


tabarra:
Uhh the trailing white space is required for a line break in the github-flavored markdown spec

cxllin2024:
I had overlooked that, but it should be fixed now.

Issue #1065, opened 2025-08-19
[BUG] monitor client-side events are missing validation


Jerrys-C:
**txAdmin/FXServer versions:**  
FXServer v8.0.1  
Resource: `monitor`  

**Describe the bug**  
Several client events are missing server-side validation. Any cheater or client-side trigger can directly execute these events without proper permission checks.  

Affected events include:  
- `txcl:spectate:start`  
- `txcl:heal`  
- `txcl:setPlayerMode`  
- `txcl:tpToWaypoint`  

**To Reproduce**  
1. Open the client console.  
2. Execute `TriggerEvent('txcl:heal', -1)` or other listed events.  
3. The event executes successfully without requiring proper authorization.  

**Expected behavior**  
These events should be validated on the server-side to ensure only authorized administrators or players with specific permissions can trigger them.  

**Screenshots**  
N/A  

**Additional context**  
The lack of validation for these events poses a serious security and gameplay integrity issue, allowing malicious players to exploit server functionality.

tabarra:
This report makes no sense.  
You say that it lacks server-side validation, but this is exactly the response to the server side validation.
Also, any cheater can do all those things even without txAdmin, so even if I "patch" this "exploit", it would not change anything because cheaters would still be able to do all of those things...

Jerrys-C:
The real issue is that you relied on client-side events, which means any client can freely trigger them with TriggerEvent without ever calling server-side validation. Saying it “wouldn’t change anything” is completely absurd—any plugin with proper security mechanisms does not suffer from this problem, because the first rule is: never trust the client.

GoatG33k:
If they already have the ability to execute within the Lua runtime the entire point is moot anyway because they can just run code to the exact same effect with or without tx events. It does have proper security mechanisms, you're just lacking the context to understand that.

Jerrys-C:
And while cheaters can always abuse native client functions, the undeniable fact here is that they currently have unrestricted access to multiple public client events exposed by txAdmin. In other words, instead of relying solely on game-native functions, they can freely exploit these txAdmin-provided events, which makes the situation significantly worse.

Gogsi:
It doesn't matter if the cheaters uss their executor to `TriggerEvent('txcl:heal', -1)` or to `SetEntityHealth(PlayerPedId(), 200)`, it's exactly the same. In fact the event just ends up calling  SetEntityHealth. 

[...]

PR #1088, opened 2025-12-15
refactor(menu): migrate permission checks to server-side callbacks

bedoflex:
Implemented a lightweight callback system to enforce server-side permission validation.

Previously, permission checks were relying on client-side data, which made the system vulnerable to exploitation via Lua injectors.

Changes:
- Added `sv_callback.lua` and `cl_callback.lua` for handling server-client requests.
- Refactored `cl_functions.lua` and menu logic to validate permissions via server callback before execution.
- Ensured all menu actions are now strictly verified by the server.

### DIFF
--- a/fxmanifest.lua
+++ b/fxmanifest.lua
@@ -24,6 +24,7 @@ shared_scripts {
 server_scripts {
     'entrypoint.js',
     'resource/sv_main.lua', --must run first
+    'resource/sv_callback.lua',
     'resource/sv_admins.lua',
     'resource/sv_logger.lua',
     'resource/sv_resources.lua',
@@ -43,6 +44,7 @@ server_scripts {
 
 client_scripts {
     'resource/cl_main.lua',
+    'resource/cl_callback.lua',
     'resource/cl_logger.lua',
     'resource/cl_playerlist.lua',
     'resource/menu/client/cl_webpipe.lua',
--- a/resource/cl_callback.lua
+++ b/resource/cl_callback.lua
@@ -5,13 +5,16 @@ function TriggerServerCallback(name, ...)
     local requestId = GetGameTimer() .. math.random(1, 999999)
     pendingCallbacks[requestId] = p
     TriggerServerEvent(eventReq, name, requestId, ...)
-    return Citizen.Await(p)
+    
+    local result = Citizen.Await(p)
+    
+    return table.unpack(result)
 end
 
 RegisterNetEvent(eventRes, function(requestId, ...)
     local p = pendingCallbacks[requestId]
     if p then
-        p:resolve(...)
+        p:resolve({ ... })
         pendingCallbacks[requestId] = nil
     end
 end)
--- a/resource/cl_callback.lua
+++ b/resource/cl_callback.lua
@@ -1,0 +1,17 @@
+local pendingCallbacks = {}
+
+function TriggerServerCallback(name, ...)
+    local p = promise.new()
+    local requestId = GetGameTimer() .. math.random(1, 999999)
+    pendingCallbacks[requestId] = p
+    TriggerServerEvent(eventReq, name, requestId, ...)
+    return Citizen.Await(p)
+end
+
+RegisterNetEvent(eventRes, function(requestId, ...)
+    local p = pendingCallbacks[requestId]
+    if p then
+        p:resolve(...)
+        pendingCallbacks[requestId] = nil
+    end
+end)
--- a/resource/menu/client/cl_base.lua
+++ b/resource/menu/client/cl_base.lua
@@ -90,11 +90,10 @@ end)
 if not TX_MENU_ENABLED then return end
 
 -- Checking with server if we are an admin
-TriggerServerEvent('txsv:checkIfAdmin')
-
--- Triggered as callback of txsv:checkIfAdmin
-RegisterNetEvent('txcl:setAdmin', function(username, perms, rejectReason)
-  if type(perms) == 'table' then
+function checkIfAdmin() 
+  local username, perms, rejectReason = TriggerServerCallback('txsv:cb:checkPerms')
+  print(username, perms, rejectReason)
+  if username and perms and type(perms) == "table" then
     debugPrint("^2[AUTH] logged in as '" .. username .. "' with perms: " .. json.encode(perms or "nil"))
     menuIsAccessible = true
     menuPermissions = perms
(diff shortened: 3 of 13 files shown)


tabarra:
This does not fix anything.  
At the end of the day, those callbacks can still be manipulated by a cheater just like events.  
  
Furthermore, all those actions (player IDs, teleport, god mode, etc.) can be done by any cheater **even if the server does not have txAdmin**, so how changing _anything_ in txAdmin is going to solve it?  
The best thing that can be done is just adding some checks to auto-kick the player, and any remotely competent cheat developer will be able to bypass that within minutes, so why bother?  
  
Please read this reply, where I tried to explain a little better: https://github.com/citizenfx/txAdmin/issues/1065#issuecomment-3201501480

bedoflex:
> Bu hiçbir şeyi düzeltmez. Sonuçta, bu geri dönüşler yine de bir hile tarafından olaylar gibi manipüle edilebilir.
> 
> Ayrıca, tüm bu işlemler (oyuncu kimlikleri, ışınlanma, tanrı modu vb.) **sunucuda txAdmin olmasa** bile herhangi bir hileci tarafından yapılabilir, peki txAdmin'de _herhangi bir_ şeyi değiştirmek bunu nasıl çözecek? Yapılabilecek en iyi şey, oyuncuyu otomatik olarak atmak için bazı kontroller eklemek ve biraz yetkin herhangi bir hile geliştiricisi bunu dakikalar içinde atlatabilir, o halde neden uğraşalım ki?
> 
> Please read this reply, where I tried to explain a little better: [#1065 (comment)](https://github.com/citizenfx/txAdmin/issues/1065#issuecomment-3201501480)

Even this small improvement will make thousands of cheaters try again, and maybe even make txAdmin stop trying, but by doing nothing, you are elevating the cheaters even higher, and they laugh at you while cheating, seeing you as a fool.

yorick2002:
> > Bu hiçbir şeyi düzeltmez. Sonuçta, bu geri dönüşler yine de bir hile tarafından olaylar gibi manipüle edilebilir.
> > Ayrıca, tüm bu işlemler (oyuncu kimlikleri, ışınlanma, tanrı modu vb.) **sunucuda txAdmin olmasa** bile herhangi bir hileci tarafından yapılabilir, peki txAdmin'de _herhangi bir_ şeyi değiştirmek bunu nasıl çözecek? Yapılabilecek en iyi şey, oyuncuyu otomatik olarak atmak için bazı kontroller eklemek ve biraz yetkin herhangi bir hile geliştiricisi bunu dakikalar içinde atlatabilir, o halde neden uğraşalım ki?
> > Please read this reply, where I tried to explain a little better: [#1065 (comment)](https://github.com/citizenfx/txAdmin/issues/1065#issuecomment-3201501480)
> 
> Even this small improvement will make thousands of cheaters try again, and maybe even make txAdmin stop trying, but by doing nothing, you are elevating the cheaters even higher, and they laugh at you while cheating, seeing you as a fool.

The only fool here is you Bedoflex

S4CKY585:
honestly there is no harm in adding security, it can be seen as a way to catch out stupid cheaters.

I totally understand the perspective of they can do the exact same thing they're exploiting with natives. however this just gives you an extra way to catch cheaters. 🤷‍♂️ (just my opinion, doesn't bother me regardless as both views are very understandable)

eventually cheaters will just not try and abuse txAdmin events if its secured and use natives instead :/ , but why not flag up stupid cheaters now whilst they are still for some reason using tx events.

I also think this is a benefit for new developers to learn how to harden or secure their events when they view txAdmin's open-source code :)

S4CKY585:
> 
> Even this small improvement will make thousands of cheaters try again, and maybe even make txAdmin stop trying, but by doing nothing, you are elevating the cheaters even higher, and they laugh at you while cheating, seeing you as a fool.

I dont see how the cheaters are being elevated and they literally are the fools for using an event instead of a native to do the same thing 😂

tabarra:
Alright, it's time to close this PR, with the reasons listed below.

## Implementation
The positive thing about the proposed changes is that adding a random ID to each event adds a layer of complexity for abusing the events, and I definitely see the value in that. But this is not enough to merge this PR.  
What this PR fundamentally does is to add a second round trip answer to check for permissions using a different-ish method of doing so (the random ID thing), but this does not fundamentally change the issue at hand, which is the fact that these actions are executed on the client.  
  
Other than the issue pointed above, the implementation could have achieved the same by just adding the extra random ID protection to the event responsible for filling in the client local permissions table (`menuPermissions`), which would have the same effect without the added complexity being spread across multiple files and the action latency issue caused by a second round trip.  

## Responding to comments  
> Even this small improvement will make thousands of cheaters try again [...]

@bedoflex Although I sympathize with this argument, the current implementation is not the way to go about it.  
  
> [...] but by doing nothing, you are elevating the cheaters even higher, and they laugh at you while cheating, seeing you as a fool.

Just because you don't see commits, that does not mean me (or the Cfx team) is "doing nothing".  
And I care about the opinion of cheaters as much as they care about ruining the game for everyone else.  
  
> Previously, permission checks were relying on client-side data, which made the system vulnerable to exploitation via Lua injectors.
> Changes: Ensured all menu actions are now strictly verified by the server.

The proposed implementation **still relies on client-side data**, and the actions were already being "strictly verified by the server".  
  
> honestly there is no harm in adding security, it can be seen as a way to catch out stupid cheaters.

@S4CKY585 Well, if this neither blocks, nor makes it more difficult for cheaters to cheat, then it's not really adding security.  
Furthermore, there is definitely harm in merging this code, both in abstract ways (eg. project maintainability), and in very concrete ways (added action latency).  
And this doesn't really "catch" stupid cheaters since it does not trigger any action like banning the cheater.
  
> I also think this is a benefit for new developers to learn how to harden or secure their events

Except they would be learning how to perform [Security Theater](https://en.wikipedia.org/wiki/Security_theater), as this is not the right way to deal with this issue.  
  
## Conclusion
**The issue at hand is not a txAdmin vulnerability**, but instead:  
1. The public perception of txAdmin's security due to misunderstanding and misinformation;
2. And that's because of a fundamental misunderstanding of security layers and the limitations of client-side validations.  

I do see the value of a _nonce_ (random ID) in events, and this added entropy is already a (small) part of the planned changes to tackle the "exploitation" problem above, and **I do appreciate the PR and attempted solution** for the problem, but I will close this PR.  
  
**I hope this answer was elucidating for everyone.**  
Please feel free to continue commenting, even if the thread is closed.

PR #1091, opened 2026-01-01
Fix FxScheduler cron job

Mathu-lmn:
Sometimes, the checkSchedule was triggered twice in the same minute (probably one by the "cron" and one by another function)
This caused the txAdmin:events:scheduledRestart event to be sent twice and ultimately (for us) the "reboot in x minutes" announcement message in our discord guild was sent 2 times


yorick2002:
Change target branch to dev please 🤓

PR #1094, opened 2026-01-18
fix(cl_vehicle): Pass correct vehicle type on spawn request

yorick2002:
This PR fixes the issue where txAdmin passes the wrong vehicle type causing the menu to spawn an oil tanker instead of the vehicle you are trying to spawn.

Thanks to Ehbw for providing the fix.

Reported in: https://discord.com/channels/192358910387159041/1462429224311718020

### DIFF
--- a/resource/menu/client/cl_vehicle.lua
+++ b/resource/menu/client/cl_vehicle.lua
@@ -22,54 +22,21 @@ local function getPedVehicle()
     end
 end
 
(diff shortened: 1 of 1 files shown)


tabarra:
Need to check with the framework team if something like `GetVehicleTypeFromName()` is already available

tabarra:
Also, related to #1025

PR #1097, opened 2026-02-12
feat(terminal) Allow interactive terminal via convar.

luxu-gg:
A very simple change that allows users to set a convar `txAdmin-interactive` to recover direct server terminal input.

Usage example:
`./FXServer.exe  +set txAdmin-interactive true`

### DIFF
--- a/core/modules/FxRunner/index.ts
+++ b/core/modules/FxRunner/index.ts
@@ -203,7 +203,8 @@ export default class FxRunner {
             fxSpawnVars.args,
             {
                 cwd: fxSpawnVars.dataPath,
-                stdio: ['pipe', 'pipe', 'pipe', 'pipe'],
+                // Check whether to allow user input in the terminal
+                stdio: GetConvar("txAdmin-interactive", "false") === "true" ? "inherit" : ['pipe', 'pipe', 'pipe', 'pipe'],
             },
         );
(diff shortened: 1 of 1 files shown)


yorick2002:
This is not possible for a reason and I doubt it will get merged. You can however just type `svgui` in the txAdmin live console if you're on Windows and it wil give you a similar thing

luxu-gg:
> This is not possible for a reason and I doubt it will get merged. You can however just type `svgui` in the txAdmin live console if you're on Windows and it wil give you a similar thing

You are correct. 
This does break txadmin console when enabled.

tabarra:
The reason this breaks txAdmin is because you are replacing every stdio with "inherit", which will become `[inherit, inherit, inherit]`, but not only you don't need to inherit all interfaces (just the first one, stdin), but it also omits the 4th IO, which is used for the server to talk to txAdmin.  
  
But even if you change just the first one to inherit, I'm pretty sure that will still not work because fxserver does _not_ pipe in it's own stdin into txAdmin's stdin.

luxu-gg:
Appreciate the explanation, I initially only had tested the terminal.
This is really an unfortunate limitation of node. 
Svgui is not available for ssh/linux users, which requires them to access the txadmin web panel.

PR #1103, opened 2026-03-15
fix(core): enhance IP address handling and formatting across multiple…

wmade:
Add compatibility and support for IPv6.
The fxserver is already compatible, and it's a shame that you can't use txadmin if you don't have a public IPv4 address. (RFC 8200)

### DIFF
--- a/core/boot/getHostVars.ts
+++ b/core/boot/getHostVars.ts
@@ -38,7 +38,7 @@ export const hostEnvVarSchemas = {
         (val) => val < 40120 || val > 40150,
         'FXS_PORT cannot be between 40120 and 40150'
     ),
-    INTERFACE: z.string().ip({ version: "v4" }),
+    INTERFACE: z.string().ip(),
 
     //Provider
     PROVIDER_NAME: z.string()
--- a/core/boot/startReadyWatcher.ts
+++ b/core/boot/startReadyWatcher.ts
@@ -8,7 +8,7 @@ import got from '@lib/got';
 import getOsDistro from '@lib/host/getOsDistro.js';
 import { txDevEnv, txEnv, txHostConfig } from '@core/globalData';
 import consoleFactory from '@lib/console';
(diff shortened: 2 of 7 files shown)


yorick2002:
> Add compatibility and support for IPv6. The fxserver is already compatible, and it's a shame that you can't use txadmin if you don't have a public IPv4 address. (RFC 8200)

txAdmin has had ipv6 support for years

wmade:
No, otherwise I wouldn't have made the changes. And the server wouldn't crash on startup.
Unless, of course, I don't start it correctly. If I don't set TXHOST_INTERFACE, the server starts in IPv6 mode but is inaccessible.
Testing on Linux, you can clearly see that it only listens on 0.0.0.0.
And it doesn't display the address correctly either, without the [].
And if I use TXHOST_INTERFACE, the server crashes immediately. The IPv6 isn't correct.

I'm in a situation where I'm IPv6-only.

yorick2002:
> No, otherwise I wouldn't have made the changes. And the server wouldn't crash on startup. Unless, of course, I don't start it correctly. If I don't set TXHOST_INTERFACE, the server starts in IPv6 mode but is inaccessible. Testing on Linux, you can clearly see that it only listens on 0.0.0.0. And it doesn't display the address correctly either, without the []. And if I use TXHOST_INTERFACE, the server crashes immediately. The IPv6 isn't correct.
> 
> I'm in a situation where I'm IPv6-only.

Strange, because I remember in 2022 Tabby was adding IPV6 support https://github.com/citizenfx/txAdmin/commit/e70b06c45a7aabfd184be1e9c13fe1789fe632eb

dop42:
Fully supporting this PR. I've seen the same behavior: txAdmin defaulting to 0.0.0.0 even when the host is IPv6. The current implementation seems to lack the proper interface binding for modern network stacks. The fix for TXHOST_INTERFACE handling is crucial for anyone running on providers that are phasing out public IPv4s.

PR #1105, opened 2026-04-02
feat: add reverse proxy support via trusted IP range config

Simon2122:
Add support for running txAdmin behind a reverse proxy by configuring trusted proxy IP ranges. When a request comes from a trusted proxy IP, the real client IP is resolved from the X-Forwarded-For header.

Configuration via convar:
  set txAdminProxyIpRange '10.0.0.0/8,172.16.0.0/12'

Or via environment variable:
  TXHOST_PROXY_IP_RANGE=10.0.0.0/8,172.16.0.0/12

Changes:
- New CIDR range parsing/matching utility (isIpInRanges)
- New proxy IP resolver that walks XFF right-to-left
- Updated ctxVarsMw, WebSocket, and rate limiter to use resolved IP
- Added PROXY_IP_RANGE to host env var schemas
- Added txAdminProxyIpRange convar support
- Unit tests for CIDR matching

### DIFF
--- a/core/boot/getHostVars.ts
+++ b/core/boot/getHostVars.ts
@@ -3,6 +3,7 @@ import { z } from "zod";
 import { fromZodError } from "zod-validation-error";
 import fatalError from '@lib/fatalError';
 import consts from '@shared/consts';
+import { proxyIpRangeSchema } from '@lib/host/isIpInRanges';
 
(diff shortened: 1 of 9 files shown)


Paathz:
This would be great!

PR #1110, opened 2026-04-11
chore(setup/panel): add missing frameworks

PsychoShock:


tabarra:
Although it's a perfectly fair wording change, I wonder if it's a good change because people do know what qbcore and esx are, but might not understand what frameworks are.
Of course, I'm mostly talking about new server owners, as experienced ones probably don't even need to read the descriptions.
🤔

PsychoShock:
> Although it's a perfectly fair wording change, I wonder if it's a good change because people do know what qbcore and esx are, but might not understand what frameworks are. Of course, I'm mostly talking about new server owners, as experienced ones probably don't even need to read the descriptions. 🤔

I would say it's not fair for others current frameworks / future one's. I do agree, but I feel whenever the user is reaching this page, he should have a bit common sens to know what "framework" means.

lLukDreams:
if they know what qbcore or esx are, they are already the ones who dont care, as someone who was new to fivem this was really misleading at the beginning

PR #1128, opened 2026-06-22
Feat/auto restart update file

Enz0Z:
## Summary

Adds an opt-in feature that lets txAdmin automatically schedule a server restart when an *update file* (default `.update`) appears in the root of the **Server Data folder** — the same folder that usually contains `server.cfg`.

This makes CI/CD-based deployments a one-step operation: after the pipeline copies the new files into place, it just drops a single file, and txAdmin handles the player-warned, graceful restart so the new files are loaded.

> Full documentation: [`docs/auto-restart-on-update-file.md`](docs/auto-restart-on-update-file.md)

## Motivation

Servers deployed via CI/CD (GitHub Actions, GitLab CI, Jenkins, deploy scripts, etc.) need a reliable way to apply a new build. Restarting the process directly skips player warnings and graceful shutdown, and calling the API from a pipeline requires authentication and extra plumbing. This reduces the whole operation to writing one file to disk.

## How it works

- On each scheduler tick (~60s), **if the feature is enabled and the server is running**, txAdmin checks for `<ServerDataFolder>/<UpdateFileName>`.
- If found: it reads the content, **deletes the file first** (so it's never processed twice), then schedules a temporary restart `+N` minutes via the existing scheduler.
- Players get the standard countdown warnings (30/15/10/5/4/3/2/1 min, in-game + Discord) and a graceful shutdown.
- If the file has content (e.g. a version or commit hash), it's used as the restart message in the logs and in-game warning.

## What's new

**Config (`restarter` scope):**

| Setting | Key | Default |
| --- | --- | --- |
| Auto-Restart on Update File | `updateFileEnabled` | `false` |
| Update File Restart Delay (min, 1–1439) | `updateFileDelay` | `2` |
| Update File Name | `updateFileName` | `.update` |

**Settings UI:** the **FXServer** tab is now split into two cards/islands (mirroring the existing **Game** tab layout): **Settings** (unchanged) and a new **CI/CD** card holding the options above.

## Safeguards

- Only acts while the server is running; otherwise the file is left to be processed once the server is back up.
- File is deleted before scheduling → no restart loops.
- Does not override an already-pending temporary restart.
- Path traversal prevented (only the file's base name is resolved inside the Server Data folder).
- No new translation strings — reuses the existing scheduled-restart messages.

## Usage example

```bash
# After deploying your resources...
echo "v1.4.2 ($(git rev-parse --short HEAD))" > /path/to/serverdata/.update
```

## Testing

- `core` and `panel` type checks pass.
- Full production build (`npm run build`) succeeds.
- Manually verified end-to-end: enabling the feature, dropping a `.update` file, and observing the warned, scheduled restart with the file content as the restart message.

### DIFF
--- a/core/modules/ConfigStore/schema/restarter.ts
+++ b/core/modules/ConfigStore/schema/restarter.ts
@@ -31,9 +31,33 @@ const resourceStartingTolerance = typeDefinedConfig({
     fixer: SYM_FIXER_DEFAULT,
 });
 
+const updateFileEnabled = typeDefinedConfig({
+    name: 'Auto-Restart on Update File',
+    default: false,
+    validator: z.boolean(),
+    fixer: SYM_FIXER_DEFAULT,
+});
+
+const updateFileName = typeDefinedConfig({
+    name: 'Update File Name',
+    default: '.update',
+    validator: z.string().trim().min(1).max(255),
+    fixer: SYM_FIXER_DEFAULT,
+});
+
+const updateFileDelay = typeDefinedConfig({
+    name: 'Update File Restart Delay',
+    default: 2, //minutes
+    validator: z.number().int().min(1).max(1439),
+    fixer: SYM_FIXER_DEFAULT,
+});
+
 
 export default {
     schedule,
     bootGracePeriod,
     resourceStartingTolerance,
+    updateFileEnabled,
+    updateFileName,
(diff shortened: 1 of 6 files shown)


yorick2002:
claude code write auto restart update file PR make no mistakes pls ty

Issue #1055, opened 2025-05-16
[FEATURE]: Player Jail or Timeout


byayex:
### Scope

Web, In-Game Menu, Developer API

### Feature Description

I would like to have the ability to jail online/offline players via the web/ingame-panel, for a specific reason and time.
An adminjail should stop all movements, stop the player from using any other scripts (if possible), teleport the player (a: to another dimension, b: to any place under the map). 

### Use Case

- Player has violated a rule, but it is too much to ban him (Ban is often not the right punishment).

### Proposed Solution

We could just include the functionality in txAdmin.

### Additional Info

If this is in the scope of this project, I would love to contribute this feature.

0-codex:
TxAdmin primarily functions as a player management interface, not a roleplay (RP) intervention tool.

byayex:
I get what you're saying, but I think you might be looking at this the wrong way. I'm not asking for a roleplay feature. This is supposed to be a tool for admins, plain and simple.

Think of it as a step between a simple "warn" and a full-on "ban." Sometimes a player messes up, but not badly enough to get banned. A kick is useless because they can just come right back. A temp-ban just means they go play somewhere else for a while.

This "jail" idea is a real punishment. It forces the player to actually stop and wait, right there in the game. It's an admin freezing a player because they broke a rule—not a cop arresting them for an in-game crime. That's why I suggested things like teleporting them under the map or to another dimension; it's a **technical action, not roleplay.**

Basically, it gives admins a better option for managing their server and dealing with rule-breakers. It's not about adding to RP, it's about giving admins more control.

Besides, there's no real disadvantage to having this feature in txAdmin.

tabarra:
Actually, I see what you mean.  
What you looking for is some kind of "timeout" action.  
  
Server owners often use the kick as a timeout (which I never really liked) and I'm considering alternatives on how to implement that.  
I see your suggestion as more sophisticated, but I'm worried that admins will still prefer to just kick the player instead.  
  
If you don't mind, I'll rename and reopen this issue, for future feedback.

tabarra:
Adding some caveats from a discord convo:  

1. Which bucket to pick? Whatever number we pick someone somewhere will be using it for some reason, so would likely need to check all buckets for an empty one and then send some event so other resources know that that bucket is busy.
2. Just keeping the player locked in is already hard… so many random resources interactions that would make the user do weird stuff (other than just teleporting, which we could counter)
[...]

PR #1131, opened 2026-07-03
feat: adding jail feature

byayex:
Adds a "jail" (timeout) action as an alternative to bans/kicks — instead of removing a player from the server, admins can now confine them in-game for a set duration.

- **Database**: new `jail` action type, with DAO methods to register a jail, add served time, and mark a sentence as finished. Served time only counts while the player is actually online, and it's clamped to the sentence duration. Jails can be revoked the same way bans/warns can.
- **Config**: added `jailRoutingBucket` (default `916`) and per-game jail coordinates (`jailPosFivem` / `jailPosRedm`, defaulting to Mission Row PD and the Valentine sheriff jail), both validated on input.
- **Enforcement**: new `jailUtils.ts` builds the "jail environment" (bucket + coords) sent to the client; wired through player classes, the player list, and FD3 message handling.
- **Client resources**: added `sv_jail.lua` and `cl_jail.lua` to actually enforce the jail in-game (routing bucket isolation + teleport).
- **UI**: new in-game jail overlay (`JailPage.tsx`, `DialogJailView.tsx`) and web panel form/tab (`JailForm.tsx`, `PlayerJailTab.tsx`) so admins can jail/release players and see jail history.
- **Events**: documented `playerJailed` (fires on apply and on re-apply after reconnect, carries bucket/coords/remaining time) and `playerJailReleased` (fires on completion or revocation) so other resources like phone/jobs scripts can react.
- **Permissions**: added a new `players.jail` permission.

## Why
Bans and kicks are often too harsh for minor infractions. Jail gives admins a lighter-touch option while still exposing enough events/state (`txAdminJailed` state bag) for other resources to hook int

### DIFF
--- a/core/lib/misc.ts
+++ b/core/lib/misc.ts
@@ -236,7 +236,7 @@ export const anyUndefined = (...args: any) => [...args].some((x) => (typeof x ==
 /**
  * Calculates expiration and duration from a ban duration string like "1 day"
  */
-export const calcExpirationFromDuration = (inputDuration: string) => {
+export const calcExpirationFromDuration = (inputDuration: string, allowMinutes = false) => {
     let expiration;
     let duration;
     if (inputDuration === 'permanent') {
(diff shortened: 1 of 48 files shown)


byayex:
PR results of #1055

Issue #1132, opened 2026-07-03
[BUG] Ban Template Dialog CSS Issue


Auxillarus:
**txAdmin/FXServer versions:**
tx: v8.1.1 | fx: b31689/Win  

**Describe the bug**
The ban template dialog's reason input is not wide enough.

**To Reproduce**
Settings > Bans > Edit Ban Templates > Add New Reason

**Expected behavior**
The input should take up the remaining width of the dialog.

**Screenshots**
<img width="530" height="310" alt="Image" src="https://github.com/user-attachments/assets/74359038-8cd7-49d7-aea9-a3b5efd021e0" />

**Additional context**
It looks like the changes to the autosize-textarea component in commit [a94cf7a](https://github.com/citizenfx/txAdmin/commit/a94cf7aa0b7ddb33f14c691d883414aad6621bcb) added a parent div which breaks the CSS when the component is in a grid.

PR #1134, opened 2026-07-12
Fix ban template reason textarea width

IThinkImGood:
## Summary
Fixes the ban template dialog reason field not spanning the expected dialog width after the autosize textarea gained a wrapper element.

## Cause
The dialog passed grid column classes directly to `AutosizeTextarea`. Since the component now renders a wrapper around the textarea, those classes no longer applied to the grid child.

## Changes
- Move the grid span classes to a wrapper around `AutosizeTextarea`
- Fix the reason label `htmlFor` to match the textarea id


IThinkImGood:
Fixes #1132

